PR #547, opened 2014-09-24
Adds ClientMergeable

### LINKED ISSUE #459: sumByKey output in storm incorrectly reports total sum on storm
sumByKey output in storm is currently problematic, because it is hiding the fact (and was before) that it is only returning the values from the online store. The offline store result is also needed. None in the _1 part of the tuple may be a lie.

The right thing is probably a ClientStore that is from [(K, BatchID), (V, BatchID)] where the input batchID is an inclusive upper-bound on the sum, and the result BatchID is the biggest non-empty BatchID <= input.

johnynek:
Will write the tests tomorrow.

closes #459


summingbird-client/src/main/scala/com/twitter/summingbird/store/ClientMergeable.scala
```diff
@@ -0,0 +1,113 @@
...
+  private val mm = new MapMonoid[K, FOpt[V]]()(fsg)
+
+  override def multiMerge[K1 <: (K, BatchID)](ks: Map[K1, V]): Map[K1, FOpt[V]] = {
+    /*
+     * We merge the smallest batch for each key first, since subsequent
+     * batches are total sums. This could be optimized more, but since
+     * mergeing two different batches for the same key is presumably rare,
+     * it is probably not worth it.
+     */
+    val batchForKey: Map[K1, V] = ks.groupBy { case (k, _) => k }
```
johnynek: good catch! Why tests didn't is a question...

johnynek:
@ianoc can we merge?

this should be totally safe. No other code touched.

johnynek:
@ianoc @isnotinvain review on this additional code (no old code changed).

PR #578, opened 2015-01-13
Pass the reducers option down to leftJoin

egonina:
Looks like we did not pass this option down to doIndependentJoin so the reducers were always set to default.


johnynek:
merge when green

PR #581, opened 2015-01-14
Lazy store init for CombinedServiceStoreFactory

egonina:
Need to call-by-name and lazy init of ClientStore for the stores to be properly initialized on remote nodes.


ianoc:
Merge when green

egonina:
reopened in #582

PR #582, opened 2015-01-14
Lazy store init for CombinedServiceStoreFactory

egonina:
Need to call-by-name and lazy init of ClientStore for the stores to be properly initialized on remote nodes.


egonina:
@ianoc, I'm reopening this in another PR due to previously broken branch, sorry for the confusion...

Issue #542, opened 2014-08-22
Have a flag to enable the DAG optimizer


jcoveney:
See #537

johnynek:
duplicate of #634

PR #587, opened 2015-02-06
Remove StripNameNodes

johnynek:
This is part of #542 to use the DagOptimizer code more.

This addresses a main issue: the names were ill-defined after optimization before. This code changes the naming to apply to the irreducible things in the graph: functions, stores, services, sources, semigroups, sinks. So, even after optimization the names can be recovered.

Tests pass, and I think we have good tests on names, so I think we should move forward.

### DIFF
--- a/summingbird-core/src/main/scala/com/twitter/summingbird/Irreducible.scala
+++ b/summingbird-core/src/main/scala/com/twitter/summingbird/Irreducible.scala
@@ -1,0 +1,36 @@
+/*
+ Copyright 2014 Twitter, Inc.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


johnynek:
By the way, this opens up the use of the dag optimizer: we can apply the function composition optimizers, and not lose the names passed by the user. This would allow us to remove this duplicated composition code from any platform.

egonina:
LGTM, a much cleaner implementation now!

johnynek:
So, that's one +1, meeting our normal rule. Will merge today unless there is a strong objection.

ianoc:
Should we wait till post rls? Not sure we've any real testing done beyond unit tests of this rls so far, so not sure we want to wait. But we should merge asap if we want it in the rls type approach

PR #588, opened 2015-02-10
Add ComposedFunction for flatMapKeys/Values

johnynek:
If we use the rule that converts flatMapKeys/flatMapValues to a normal flatMap, we were hiding the inner functions so their names would be lost.

This does not effect using the DagOptimizer to remove the names, since we did not apply any other rules in that code.


johnynek:
@pankajroark Can we merge this? I never got a review on it.

pankajroark:
Are there any tests for the two new rules added here? If not, should we add them?

The change looks good otherwise.

johnynek:
@pankajroark I have added some tests to exercise this. Thanks for keeping the testing bar high!

pankajroark:
LGTM feel free to merge.

PR #590, opened 2015-02-11
[Don't Merge This] -- Show merge conflicts between 0.6.0 and master

isnotinvain:


project/Build.scala
```diff
@@ -46,9 +46,15 @@ object SummingbirdBuild extends Build {
 
   val sharedSettings = extraSettings ++ Seq(
     organization := "com.twitter",
+<<<<<<< HEAD
```
isnotinvain: Everything in this file is easy to fix, so no worries here

summingbird-core-java/src/main/java/com/twitter/summingbird/memory/javaapi/JMemory.java
```diff
@@ -82,6 +82,10 @@ public Void apply(IN v) {
     return new Service<Memory, JMemoryService<K, V>, K, V>(new JMemoryService(service));
   }
 
+<<<<<<< HEAD
+=======
+  private Memory platform = new Memory(new JobId("inMem"));
```
isnotinvain: was this deleted on develop on purpose, or added in master and we want to keep it?

summingbird-core/src/main/scala/com/twitter/summingbird/Stats.scala
```diff
@@ -93,11 +93,16 @@ object JobCounters {
   private val registeredCountersForJob: ConcurrentHashMap[JobId, ParHashSet[(Group, Name)]] =
     new ConcurrentHashMap[JobId, ParHashSet[(Group, Name)]]()
 
+<<<<<<< HEAD
```
isnotinvain: Assuming we just want the HEAD block here? (head is the release branch)

summingbird-core/src/main/scala/com/twitter/summingbird/memory/Memory.scala
```diff
@@ -41,8 +47,15 @@ class Memory(implicit jobID: JobId = JobId("default.memory.jobId")) extends Plat
   private type Prod[T] = Producer[Memory, T]
   private type JamfMap = Map[Prod[_], Stream[_]]
 
+<<<<<<< HEAD
```
isnotinvain: assuming HEAD is more up to date here too?

summingbird-core-java/src/main/java/com/twitter/summingbird/memory/javaapi/JMemory.java
johnynek: It was added, if I'm not mistaken.

johnynek: Wait, actually, I don't know.

johnynek: @egonina should comment here. I guess we are going to want it.

summingbird-core/src/main/scala/com/twitter/summingbird/memory/Memory.scala
egonina: Yes, HEAD is the latest

summingbird-core/src/main/scala/com/twitter/summingbird/Stats.scala
egonina: Yes

summingbird-core-java/src/main/java/com/twitter/summingbird/memory/javaapi/JMemory.java
egonina: No, we fixed it with [this](https://github.com/twitter/summingbird/blob/develop/summingbird-core-java/src/main/java/com/twitter/summingbird/memory/javaapi/JMemory.java#L33) in the later version. We can take the HEAD version here.

Issue #574, opened 2014-12-18
remove unneeded .as[] calls


johnynek:
https://github.com/twitter/summingbird/blob/develop/summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/batch/BatchedOperations.scala#L36

We refactored from DateRange to Interval[TimeStamp] at one point. There are some no-op corversions around.

PR #595, opened 2015-02-15
remove as call in BatchedOperations

adamkozuch:


adamkozuch:
Issue #574.

adamkozuch:
Should I refactor from DataRange for instance this or just leave it? https://github.com/twitter/summingbird/blob/f04e6c90849c3fa177cfdec3a7174c5aa02251f4/summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/ScaldingPlatform.scala#L106-L107

johnynek:
Let's leave the intersect method. DateRange does not have it, but we need it here.

johnynek:
Also, sorry the tests are broken now. Let's let #594 get merged, then merge develop here and see that the tests are green, then we'll merge.

johnynek:
Can you merge with develop and see if this goes green?

johnynek:
Thank you!

Issue #597, opened 2015-02-17
Re-enable tests commented out


johnynek:
some tests in #594 were commented to get it green. We need to make sure those tests are reenabled and find out why they are red (and how we didn't catch that until now).

johnynek:
#599 is the latest version of this.

PR #598, opened 2015-02-17
Re-enable failing tests

MansurAshraf:
@johnynek this is a clean PR with only failing tests for https://github.com/twitter/summingbird/issues/597


MansurAshraf:
see https://github.com/twitter/summingbird/pull/602

PR #601, opened 2015-02-20
Fixes the types off the summingbird java storm api

ianoc:
Seems like incomplete tests here in the repo


johnynek:
I guess that's correct, but I agree a test would be great here. If you need to merge to unblock someone, that's fine. Adding an issue.

PR #602, opened 2015-02-20
Reenable tests

johnynek:
These were just failing in #599 because we updated scalacheck and it was using boundaries more, and the laws were too strict.

No real breakage.


johnynek:
#597

ianoc:
LGTM, merge when green

Issue #513, opened 2014-05-21
Restate the core spark types in terms of a Reader monad


isnotinvain:
The core spark types that need access to a SparkContext should be refactored in terms of a Reader monad

Issue #604, opened 2015-02-21
ConcurrentMemory laws are flakey


johnynek:
[info] The ConcurrentMemory Platform should
[info] + singleStep w/ Int, Int, Set[Int]
[info] + singleStep w/ Int, String, List[Int]
[info] + singleStep w/ String, Short, Map[Set[Int], Long]
[info] + diamond w/ Int, Int, Set[Int]
[error] x diamond w/ String, Short, Map[Set[Int], Long]
[error]    the value is false (ConcurrentMemoryLaws.scala:157)

Usually passes. Sometimes it doesn't. May be a race condition, may be a real bug.

PR #607, opened 2015-03-01
Spark monad

xudifsd:
@isnotinvain #513 
also bumped up spark version, 0.9.0-incubating is too old

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -42,7 +42,6 @@ object SummingbirdBuild extends Build {
   val specs2Version = "1.13"
 
   val sparkCoreVersion ="1.2.0"
-  val scalazVersion = "7.1.1"
   val commonsHttpClientVersion = "3.1"
 
   val extraSettings = Project.defaultSettings ++ mimaDefaultSettings ++ scalariformSettings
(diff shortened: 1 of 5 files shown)


project/Build.scala
```diff
@@ -41,7 +41,8 @@ object SummingbirdBuild extends Build {
   val novocodeJunitVersion = "0.10"
   val specs2Version = "1.13"
 
-  val sparkCoreVersion ="0.9.0-incubating"
+  val sparkCoreVersion ="1.2.0"
+  val scalazVersion = "7.1.1"
```
johnynek: please use the Algebird implementation of this since we already depend on Algebird.

com.twitter.algebird.monad.Reader and the code may just compile.

johnynek:
Thanks for doing this!

Hopefully we can get the spark platform be as well supported as the scalding platform.

xudifsd:
@johnynek done, but I'm not sure why build failed, it's strange that it failed on summingbird-core, not summingbird-spark

johnynek:
You got bit by #604. Restarted it. We need to figure out why that test fails occasionally. Is it a real issue with that platform or is it a deeper problem that the job itself is non-determinisitic in online summingbird fundamentally?

When this is green we can merge.

xudifsd:
failed again..LOL, how to restart it?

PR #608, opened 2015-03-01
Replace deprecated ClassManifest and erasure 

adamkozuch:
ClassManifest is deprecated and it should be replaced by ClassTag. I read in documentation that replacement for erasure is now runtimeClass. Is that correct?


ianoc:
Not binary compatible but given its only used inside SB, seems good to me

ianoc:
Thanks for fixing the broken window!

PR #609, opened 2015-03-10
replace deprecated erasure by runtimeClass

adamkozuch:


johnynek:
Thanks for the cleanup. I guess we should really use ClassTag here, but that will be a binary change.

PR #610, opened 2015-03-17
Revert "Remove StripNameNodes"

MansurAshraf:
This reverts twitter/summingbird#587 in case we cant resolve namenode bug in time.


ianoc:
This internally caused an issue with options not being applied to nodes for parallelism.

ianoc:
merge when green

PR #611, opened 2015-03-17
DO not merge::: downgraded scalacheck to 1.10.0

MansurAshraf:


johnynek:
closing this for now. Reopen if this is needed (but I think we actually want to upgrade if we can. Some recent versions got some nice fixes).

PR #616, opened 2015-04-06
MergeableStoreFactory extends java.io.Serializable

egonina:
This was breaking serialization internally, adding fix


johnynek:
sorry we missed this. Generally all our traits need this. :/ Merge when green.

egonina:
Had to exclude two targets from summingbird-example to get this to pass, but the build is green now

johnynek:
I guess we can merge this. It may hit someone externally later if they don't explicitly add this stuff back, perhaps.

PR #617, opened 2015-04-09
Spout storm metrics foreach

egonina:
This was a bug where the metrics that are passed as NamedOptions weren't registered because .map on a TraversableOnce is lazy. This fixes the issue with a .foreach instead.


johnynek:
merge when green.

PR #621, opened 2015-07-29
Ianoc/move to scala test211

ianoc:
Killing all unused targets to make my life easier and then updating everything else


johnynek:
So, what was the pain in keeping the summingbird-java and spark?

This is quite a big diff with a lot going on. For instance, the sbt script update could have been done independently, also, I think the version updates. Is there any way we could break this into 3-4 PRs? It looks like you have most of the diffs as idependent changes above.

ianoc:
They are unused with questionable if its worth any effort what-so-ever to keep alive. We've deleted all references to summingbird-java internally so twitter at least doesn't even use this code at all. No one has ever used summingbird-spark.

Can split it into multiple PR's tho

rubanm:
LGTM

PR #624, opened 2015-07-31
Killing off Summingbird Java

ianoc:
We've killed all hooks internal to twitter as no one was using this. It seems far too hard to use in its current state to be practical.

Killing for code cleanup


isnotinvain:
+1

PR #625, opened 2015-08-13
Resilient to a store being ahead

ianoc:
This modifies the planning phase in writeLast such that if the version in the VKS exists we don't include it in the plan. But we continue on regardless.

This will recover much more gracefully from a job terminating in the middle with some stores written to and others not.

If all stores are written to this will still fail as the FlowDef generated will be invalid with no sink taps

### DIFF
--- a/summingbird-scalding-test/src/test/scala/com/twitter/summingbird/scalding/VersionBatchedStoreTest.scala
+++ b/summingbird-scalding-test/src/test/scala/com/twitter/summingbird/scalding/VersionBatchedStoreTest.scala
@@ -1,0 +1,159 @@
+/*
+ Copyright 2013 Twitter, Inc.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+package com.twitter.summingbird.scalding
+
+import com.twitter.algebird.{ MapAlgebra, Monoid, Group, Interval, Last }
+import com.twitter.algebird.monad._
+import com.twitter.summingbird.{ Producer, TimeExtractor, TestGraphs }
+import com.twitter.summingbird.batch._
+import com.twitter.summingbird.batch.state.HDFSState
+import com.twitter.summingbird.option.JobId
+import com.twitter.summingbird.SummingbirdRuntimeStats
+import com.twitter.summingbird.scalding.store.{ VersionedBatchStore, InitialBatchedStore }
+import com.twitter.bijection._
+import com.twitter.scalding.commons.source.VersionedKeyValSource
+
+import java.util.TimeZone
+import java.io.File
+
+import com.twitter.scalding.{ Source => ScaldingSource, Test => TestMode, _ }
+import com.twitter.scalding.typed.TypedSink
+
+import org.scalacheck._
+import org.scalacheck.Prop._
+import org.scalacheck.Properties
+
+import org.apache.hadoop.conf.Configuration
+
+import scala.collection.JavaConverters._
+import scala.collection.mutable.{ ArrayBuffer, Buffer, HashMap => MutableHashMap, Map => MutableMap, SynchronizedBuffer, SynchronizedMap }
+import scala.util.{ Try => ScalaTry }
+
+import cascading.scheme.local.{ TextDelimited => CLTextDelimited }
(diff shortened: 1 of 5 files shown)


summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/store/VersionedBatchStore.scala
```diff
@@ -200,13 +203,20 @@ class VersionedBatchStore[K, V, K2, V2](rootPath: String, versionsToKeep: Int, o
...
-        versionsToKeep = versionsToKeep))
+    val target = VersionedKeyValSource[K2, V2](rootPath,
+      sourceVersion = None,
+      sinkVersion = newVersion,
+      maxFailures = 0,
+      versionsToKeep = versionsToKeep)
+    if (!target.sinkExists(mode)) {
+      logger.info(s"Versioned batched store version for $this @ $newVersion doesn't exist. Will write out.")
+      lastVals.map(pack(batchID, _))
+        .toPipe((0, 1))
```
johnynek: I don't think we need the `.toPipe` anymore. That bug was fixed.

johnynek: Also, the `import Dsl._` above I think will be unneeded at that point.

summingbird-scalding-test/src/test/scala/com/twitter/summingbird/scalding/VersionBatchedStoreTest.scala
```diff
@@ -0,0 +1,159 @@
...
+
+      // Now Stores C & B
+      {
+        val tail = TestGraphs.multipleSummerJob[Scalding, (Long, Int), Int, Int, Int, Int, Int](source, testStoreC, testStoreB)({ t => fnA(t._2) }, fnB, fnC)
+        val scald = Scalding("scalaCheckMultipleSumJob")
+        val ws = new LoopState(intr)
+        scald.run(ws, mode, scald.plan(tail))
+      }
+
+      // Now check that the inMemory ==
```
johnynek: incomplete comment.

johnynek:
What if there is a single store? In that case, I guess there will be nothing added to the flowDef (I think). In which case, I think cascading will error, I think. Can we test this by doing a similar test with only one store?

We could test that the flowDef is nonEmpty, and if it is empty just log a warning that the WaitingState seems out of date with the stores, but mark the period of time complete.

summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/store/VersionedBatchStore.scala
```diff
@@ -200,13 +201,19 @@ class VersionedBatchStore[K, V, K2, V2](rootPath: String, versionsToKeep: Int, o
...
+      sinkVersion = newVersion,
+      maxFailures = 0,
+      versionsToKeep = versionsToKeep)
+    if (!target.sinkExists(mode)) {
+      logger.info(s"Versioned batched store version for $this @ $newVersion doesn't exist. Will write out.")
+      lastVals.map(pack(batchID, _))
+        .write(target)
+    } else {
+      logger.warn(s"""Versioned batched store version for $this @ $newVersion already exists! Will skip adding to plan.
+        | Note you must have at least one store that requires writing out or the job will fail.""".stripMargin('|'))
```
johnynek: this constraint is probably a pain point. So I think the real bug comes from crashed that happen after writing the store output, but before we commit to the WaitingState that we have processed a block of time.

ianoc: That period should be small? And your WS needs to be unreliable for this to be a thing right? It should be the small thing at the end. The only other alternative I think is to probably delete the stores when we start -- but they could be already consumed by downstream jobs on a cluster.

My understanding is the big issue is that you have a big chain of MR jobs, we finish some stores earlier than others, so if any MR job fails for any reason your sort of broken.

ianoc:
We can do that with the waiting state i suppose, is that sane/good to do? It seems like its probably ok to do. But I'm trying to think of cases where thats broken or does something unexpected.... any bad cases you can think of?

summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/store/VersionedBatchStore.scala
johnynek: How will it recover? Imagine we write the store but crash before we write to statebird. Statebird keeps telling you to run yesterday's block, but the job keeps failing because it has already written and there is only one store.

johnynek:
I was thinking that we could just add logic into `toFlow`, which I think is the only place where we build a flow.

In there, we could check if the `FlowDef` is empty, right?

johnynek:
Or, maybe not empty, because maybe by then we will have added the sources, but maybe there are no sinks. So, in that case, rather than just hand it to cascading and have it throw, we warn and report that the interval of time is finished (the data fairy left the data without us having to compute it, for all we know).

PR #626, opened 2015-08-13
Fixes tests

ianoc:
The usual it seems scalatest conversion issues. Doesn't run things that are object's vs classes.


johnynek:
jesus. Changing things about testing is so dangerous. We have no tests that the tests are testing what we think they are.

ianoc:
Yeah, a very meta problem. I think what we really want but i haven't seen travis or something do it is just something that graphs the number of tests we run. Should fail the build if it goes down without some exception i guess...

ianoc:
@johnynek well this is green now. Well know the tests are back working when the storm flaky tests means it takes several attempts to get a green run :/

PR #628, opened 2015-10-19
Ianoc/optimization changes

ianoc:
Few relatively minor changes to improve some SB performance in scalding jobs

### DIFF
--- a/summingbird-batch/src/main/scala/com/twitter/summingbird/batch/BatchID.scala
+++ b/summingbird-batch/src/main/scala/com/twitter/summingbird/batch/BatchID.scala
@@ -139,5 +139,5 @@ case class BatchID(id: Long) extends Ordered[BatchID] with java.io.Serializable 
   def min(b: BatchID) = if (this < b) this else b
 
   override lazy val toString = "BatchID." + id.toString
-  override lazy val hashCode: Int = super.hashCode
+  override lazy val hashCode: Int = id.hashCode
 }
--- a/summingbird-batch/src/main/scala/com/twitter/summingbird/batch/BatchID.scala
+++ b/summingbird-batch/src/main/scala/com/twitter/summingbird/batch/BatchID.scala
@@ -48,7 +48,6 @@ import scala.collection.Iterator.iterate
 object BatchID {
   implicit val equiv: Equiv[BatchID] = Equiv.by(_.id)
 
-  def apply(long: Long) = new BatchID(long)
   def apply(ts: Timestamp) = new BatchID(ts.milliSinceEpoch)
 
   // Enables BatchID(someBatchID.toString) roundtripping
@@ -124,22 +123,21 @@ object BatchID {
 
   implicit val batchID2Bytes: Injection[BatchID, Array[Byte]] =
     Injection.connect[BatchID, Long, Array[Byte]]
+
+  implicit val ord: Ordering[BatchID] = Ordering.by(_.id)
 }
 
-class BatchID(val id: Long) extends Ordered[BatchID] with java.io.Serializable {
+case class BatchID(id: Long) extends Ordered[BatchID] with java.io.Serializable {
   def next: BatchID = new BatchID(id + 1)
   def prev: BatchID = new BatchID(id - 1)
   def +(cnt: Long) = new BatchID(id + cnt)
   def -(cnt: Long) = new BatchID(id - cnt)
+
   def compare(b: BatchID) = id.compareTo(b.id)
+
   def max(b: BatchID) = if (this >= b) this else b
   def min(b: BatchID) = if (this < b) this else b
-  override lazy val toString = "BatchID." + id.toString
 
-  override def hashCode: Int = id.hashCode
-  override def equals(other: Any): Boolean =
-    other match {
-      case that: BatchID => this.id == that.id
-      case _ => false
-    }
+  override lazy val toString = "BatchID." + id.toString
+  override lazy val hashCode: Int = super.hashCode
 }
--- a/summingbird-batch/src/main/scala/com/twitter/summingbird/batch/BatchID.scala
+++ b/summingbird-batch/src/main/scala/com/twitter/summingbird/batch/BatchID.scala
@@ -139,5 +139,5 @@ case class BatchID(id: Long) extends Ordered[BatchID] with java.io.Serializable 
   def min(b: BatchID) = if (this < b) this else b
 
   override lazy val toString = "BatchID." + id.toString
-  override lazy val hashCode: Int = id.hashCode
+  override def hashCode: Int = id.hashCode
 }
--- a/summingbird-batch/src/main/scala/com/twitter/summingbird/batch/TimeStamp.scala
+++ b/summingbird-batch/src/main/scala/com/twitter/summingbird/batch/TimeStamp.scala
@@ -36,7 +36,7 @@ case class Timestamp(milliSinceEpoch: Long) extends Ordered[Timestamp] {
   def incrementMinutes(minutes: Long) = Timestamp(milliSinceEpoch + (minutes * 1000 * 60))
   def incrementHours(hours: Long) = Timestamp(milliSinceEpoch + (hours * 1000 * 60 * 60))
   def incrementDays(days: Long) = Timestamp(milliSinceEpoch + (days * 1000 * 60 * 60 * 24))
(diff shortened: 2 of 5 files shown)


summingbird-batch/src/main/scala/com/twitter/summingbird/batch/BatchID.scala
```diff
@@ -124,22 +123,21 @@ object BatchID {
...
-  override lazy val toString = "BatchID." + id.toString
 
-  override def hashCode: Int = id.hashCode
-  override def equals(other: Any): Boolean =
-    other match {
-      case that: BatchID => this.id == that.id
-      case _ => false
-    }
+  override lazy val toString = "BatchID." + id.toString
+  override lazy val hashCode: Int = id.hashCode
```
johnynek: does this need to be a lazy val? Maybe just val and not have the thread sync costs? hashCode of a Long must be really fast.

summingbird-batch/src/main/scala/com/twitter/summingbird/batch/TimeStamp.scala
```diff
@@ -36,6 +36,7 @@ case class Timestamp(milliSinceEpoch: Long) extends Ordered[Timestamp] {
   def incrementMinutes(minutes: Long) = Timestamp(milliSinceEpoch + (minutes * 1000 * 60))
   def incrementHours(hours: Long) = Timestamp(milliSinceEpoch + (hours * 1000 * 60 * 60))
   def incrementDays(days: Long) = Timestamp(milliSinceEpoch + (days * 1000 * 60 * 60 * 24))
+  override lazy val hashCode: Int = milliSinceEpoch.hashCode
```
johnynek: maybe just val here?

summingbird-chill/src/main/scala/com/twitter/summingbird/chill/SBChillRegistrar.scala
```diff
@@ -56,6 +56,7 @@ object SBChillRegistrar {
       classOf[ScalaKryoInstantiator],
       new ScalaKryoInstantiator()
         .withRegistrar(new IterableRegistrar(iterableRegistrars ++ defaults))
+        .setReferences(false)
```
johnynek: good catch.

summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/batch/BatchedStore.scala
```diff
@@ -33,6 +33,18 @@ import cascading.flow.FlowDef
 import org.slf4j.LoggerFactory
 
 import StateWithError.{ getState, putState, fromEither }
+import com.twitter.scalding.serialization.macros.impl.BinaryOrdering
+
+private[batch] case class LTuple2[T, U](_1: T, _2: U) {
```
johnynek: can you comment on the intent here?

summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/batch/BatchedStore.scala
```diff
@@ -93,14 +105,14 @@ trait BatchedStore[K, V] extends scalding.Store[K, V] { self =>
 
   protected def sumByBatches[K1, V: Semigroup](ins: TypedPipe[(Timestamp, (K1, V))],
     capturedBatcher: Batcher,
-    commutativity: Commutativity): TypedPipe[((K1, BatchID), (Timestamp, V))] = {
+    commutativity: Commutativity): TypedPipe[(LTuple2[K1, BatchID], (Timestamp, V))] = {
```
johnynek: this is weird. You have a protected method returning a private[batch] type. So, doesn't that effectively make this method `private[batch]` which may not be what we want, but maybe?

summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/batch/BatchedStore.scala
```diff
@@ -93,14 +105,14 @@ trait BatchedStore[K, V] extends scalding.Store[K, V] { self =>
...
     implicit val timeValueSemigroup: Semigroup[(Timestamp, V)] =
       IteratorSums.optimizedPairSemigroup[Timestamp, V](1000)
 
     val inits = ins.map {
       case (t, (k, v)) =>
         val batch = capturedBatcher.batchOf(t)
-        ((k, batch), (t, v))
+        (LTuple2(k, batch), (t, v))
```
johnynek: this is going to make it so that each key, batch pair is unique in the map, so the values will never be reduced, right? Why not just skip mapside aggregation entirely?

johnynek: This is super confusing. It seems like you are using a type to disable sumByBatches doing anything, but why not just not call it?

summingbird-batch/src/main/scala/com/twitter/summingbird/batch/TimeStamp.scala
ianoc: Far, habbit. Indeed 1234L.hashCode == 1234L on the jvm

summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/batch/BatchedStore.scala
ianoc: Sure, any concerns about doing it. Needs a comment tho

In general should we add a set of these Tuples to algebird/bijection at some point maybe?

summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/batch/BatchedStore.scala
ianoc: Your right, it doesn't make any sense, the type leaks out. I could make the tuple type public itself maybe. Any changes to this will break the public api alas, though not in a way almost anyone would ever use, but still possible.

summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/batch/BatchedStore.scala
ianoc: Can we not have multiple matching ones here? we have a monoid on Timestamp for max and a semigroup on the users V. Does this not do our normal mapside agg rather than aiming for uniqueness?

ianoc: I'm not sure what you mean here? are you talking about old behavior or new? This code change shouldn't have changed the behavior of this code, just replaced the built in tuple2 with the lazy hash code version

johnynek: I was confused. I thought you were using `System.identityHashCode`. I didn't read carefully enough. The comment helps. Never mind.

summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/batch/BatchedStore.scala
```diff
@@ -33,6 +33,23 @@ import cascading.flow.FlowDef
...
+import com.twitter.scalding.serialization.macros.impl.BinaryOrdering
+
+/**
+  * This is the same as scala's Tuple2, except the hashCode is a lazy val.
+  * We do this so when the tuple2 is placed in the map we won't caculate the hash code
+  * unnecessarily several times as the map grows or is transformed.
+  */
+case class LTuple2[T, U](_1: T, _2: U) {
+
+  override lazy val hashCode: Int = scala.runtime.ScalaRunTime._hashCode(this)
```
johnynek: looks like we are always going to hit the hashCode here. Maybe drop lazy and just directly val it?

summingbird-chill/src/main/scala/com/twitter/summingbird/chill/SBChillRegistrar.scala
ianoc: Xprof reading pass time pays off :)

summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/batch/BatchedStore.scala
ianoc: Sure yeah makes sense

PR #629, opened 2015-10-20
Removing internal config setup from scalding platform

ianoc:
1) Both Timestamp and BatchID become any val classes to erase them at runtime
 ---> Knock on effect is to have them not extend ordered. Provide an implicit enrichment to add those methods back using the ordering on the companion object
2) Remove the internal config handling from the scalding platform, delegating to the Scalding Config class instead.


summingbird-builder/src/main/scala/com/twitter/summingbird/scalding/ScaldingEnv.scala
```diff
@@ -116,7 +116,7 @@ case class ScaldingEnv(override val jobName: String, inargs: Array[String])
     val scald = Scalding(name, opts)
       .withRegistrars(ajob.registrars ++ builder.registrar.getRegistrars.asScala)
       .withConfigUpdater { c =>
-        c.updated(ajob.transformConfig(c.toMap))
+        com.twitter.scalding.Config.tryFrom(ajob.transformConfig(c.toMap).toMap).get
```
johnynek: maybe import Config here? Looks a little strange.

summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/ScaldingPlatform.scala
```diff
@@ -652,67 +652,45 @@ class Scalding(
...
-      System.currentTimeMillis.toString)
-    conf.set("summingbird.submitted.timestamp",
-      System.currentTimeMillis.toString)
-
-    def ifUnset(k: String, v: String) { if (null == conf.get(k)) { conf.set(k, v) } }
-    // Set the mapside cache size, this is important to not be too small
-    ifUnset("cascading.aggregateby.threshold", "100000")
+    val postConfig = config.+("summingbird.options" -> options.toString)
+      .+("summingbird.jobname" -> jobName)
+      // legacy name to match scalding
```
johnynek: I don't think this is needed anymore (I think scalding already puts in the scalding.flow.submitted.timestamp).

ianoc: Should we keep the SB one? or just drop both?

summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/ScaldingPlatform.scala
```diff
@@ -652,67 +652,43 @@ class Scalding(
...
-    Scalding.logger.debug("Removes: {}", transformedConfig.removes)
-    Scalding.logger.debug("Updates: {}", transformedConfig.updates)
+    val initKryo = conf.getKryo match {
+      case None =>
+        new serialization.KryoHadoop(ScalaMapConfig(conf.toMap))
+      case Some(kryo) => kryo
+    }
+    conf
+      .setSerialization(
+        Left((classOf[serialization.KryoHadoop], initKryo.withRegistrar(kryoReg))), Nil)
```
johnynek: can we use the `.serSerialization(Right(...))` case if passedRegistrars.isEmpty?

summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/store/DirectoryBatchedStore.scala
```diff
@@ -81,7 +81,7 @@ class DirectoryBatchedStore[K <: Writable, V <: Writable](val rootPath: String)(
 
         val lastBatchStatus =
           hdfsPaths.map(getFileStatus(_, conf))
-            .filter { input => input._1 && (BatchID(input._2) < exclusiveUB) }
+            .filter { input => input._1 && (BatchID(input._2.milliSinceEpoch) < exclusiveUB) }
```
johnynek: this does not seem correct to me. You shouldn't just pass a millis to a `BatchID`. That's what Batchers do. Why is this change here?

ianoc: Thats what I thought too, but actually this is what we were doing here before too. 

BatchID's constructor that I removed as part of this (had to for AnyVal erasure reasons):
 def apply(ts: Timestamp) = new BatchID(ts.milliSinceEpoch)

So this prior code here was actually doing semantically the same as this new code

johnynek: Looks like @reconditesea wrote this code. It seems like it should be `batcher.batchOf(input._2)` and while we are at it, I'd rewrite this as:

```scala
.filter { case (isGood, timestamp, _) => isGood && (batcher.batchOf(timestamp) < exclusiveUB) }
```
I guess there are no tests on this code. :/

johnynek: by the way, that `apply(ts: Timestamp)` just seems like a broken method. How did that stick around so long?

ianoc: Thats a good question, i guess we don't look at this stuff that often

reconditesea:
uh, this hurts :( 
But yeah, I think we should just remove this file. Even I cannot quite remember why I/we want to add this DirectoryBatched store in the first place.

johnynek:
Looks good to me. I'm slightly worried about the source incompatibility here, but the point is better control over how the job is run and that is really important and not a small issue.

ianoc:
Awesome, many thanks for the review help!

PR #632, opened 2015-11-13
Fixes a bug in the LTuple2 class. Add a test for it

ianoc:


summingbird-scalding-test/src/test/scala/com/twitter/summingbird/scalding/batch/LTuple2Test.scala
```diff
@@ -0,0 +1,89 @@
...
+      val ltup2 = LTuple2(a2, b2)
+
+      if (a1 == a2 && b1 == b2)
+        ltup1 == ltup2
+      else
+        ltup1 != ltup2
+    }
+  }
+
+  property("Hash code of the LTuple2 is the same as a tuple2 would have been") = {
```
rubanm: Just curious how this hold true given LTuple2 and Tuple2 don't have any type hierarchy between them. This will break if the ScalaRunTime implementation changes in the future?

rubanm:
LGTM, merge when green

summingbird-scalding-test/src/test/scala/com/twitter/summingbird/scalding/batch/LTuple2Test.scala
ianoc: Both LTuple2 and Tuple2 are case classes, and both use the case class hash code method. So they should be synchronized across any upgrades. If something happens and that doesn't hold then its probably good we look at this for this test failing

johnynek:
:+1:

Issue #605, opened 2015-02-21
StormPlatform laws are flakey


johnynek:
[error] x StormPlatform matches Scala for single step jobs
[error]  the value is false (StormLaws.scala:132)

Usually passes. Sometimes it doesn't. May be a race condition, may be a real bug.

egonina:
+1

johnynek:
at least getting a clearer error report would help. :/

johnynek:
I think it may help to use a more modern storm.

@pankajroark any news on getting heron to use the modern storm namespaces so we can use storm 1.0 or so and see if we can fix our issues?

PR #637, opened 2015-11-22
re-enable Dag Optimizer to strip names

johnynek:
This reverts #610 except with simpler, and much better tested code (which is to say, actual tests and tests which exposed bugs).

This adds more tests, for e.g. #633

This does not yet enable turning on all the optimizations + naming, as you have to be more careful about "irreducibles", which can change under some optimizations (namely, composing functions). But it is enough to remove the code duplication between StripNameNodes and the DagOptimizer which can do the same thing in a type-safe and general way.

I'd like to get this merged, then tackle keeping names correct under general optimizations, which I think is possible. This will dramatically simplify implementing platforms if we can successfully get the optimizations + names working correctly.

Of course, the names could control potentially optimizations (since named options can be options for optimizations), but we can deal with that down the line (probably, by passing the options into the dag optimizer).

### DIFF
--- a/summingbird-online/src/main/scala/com/twitter/summingbird/planner/StripNamedNodes.scala
+++ b/summingbird-online/src/main/scala/com/twitter/summingbird/planner/StripNamedNodes.scala
@@ -25,7 +25,7 @@ case class ProducerF[P <: Platform[P]](oldSources: List[Producer[P, Any]],
 
 object StripNamedNode {
 
-  private def castTail[P <: Platform[P], T](node: Producer[P, T]): TailProducer[P, T] =
+  private[this] def castTail[P <: Platform[P], T](node: Producer[P, T]): TailProducer[P, T] =
     node.asInstanceOf[TailProducer[P, T]]
 
   /**
@@ -42,7 +42,12 @@ object StripNamedNode {
       case LeftJoinedProducer(_, serv) => Some(serv)
       case Summer(_, store, semi) => Some((store, semi))
       case WrittenProducer(_, sink) => Some(sink)
-      case _ => None
+      // The following have nothing to put options on:
+      case AlsoProducer(_, producer) => None
+      case NamedProducer(producer, _) => None
+      case IdentityKeyedProducer(producer) => None
+      case MergedProducer(l, r) => None
+      case _ => sys.error("Unreachable. Here to warn us if we add Producer subclasses but forget to update this")
     }
 
   def apply[P <: Platform[P], T](tail: TailProducer[P, T]): (Map[Producer[P, Any], List[String]], TailProducer[P, T]) = {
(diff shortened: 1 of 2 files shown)


summingbird-online/src/main/scala/com/twitter/summingbird/planner/StripNamedNodes.scala
```diff
@@ -24,147 +25,67 @@ case class ProducerF[P <: Platform[P]](oldSources: List[Producer[P, Any]],
 
 object StripNamedNode {
 
-  def castTail[P <: Platform[P]](node: Producer[P, Any]): TailProducer[P, Any] = node.asInstanceOf[TailProducer[P, Any]]
-  def castToPair[P <: Platform[P]](node: Producer[P, Any]): Producer[P, (Any, Any)] = node.asInstanceOf[Producer[P, (Any, Any)]]
+  private def castTail[P <: Platform[P], T](node: Producer[P, T]): TailProducer[P, T] =
```
ianoc: I don't think we've a companion class instance here, if thats the case private[this] is better for performance

summingbird-online/src/main/scala/com/twitter/summingbird/planner/StripNamedNodes.scala
```diff
@@ -24,147 +25,67 @@ case class ProducerF[P <: Platform[P]](oldSources: List[Producer[P, Any]],
...
-        processLevel(optLast, v, curMap, op)
+      case Source(src) => Some(src)
+      case OptionMappedProducer(_, fn) => Some(fn)
+      case FlatMappedProducer(_, fn) => Some(fn)
+      case ValueFlatMappedProducer(_, fn) => Some(fn)
+      case KeyFlatMappedProducer(_, fn) => Some(fn)
+      case LeftJoinedProducer(_, serv) => Some(serv)
+      case Summer(_, store, semi) => Some((store, semi))
+      case WrittenProducer(_, sink) => Some(sink)
+      case _ => None
```
ianoc: I'm a little concerned about this in that we must ensure this is an exhaustive list here, if any new elements of the graph get added (like the addition of the ValueFlatMappedProducer), we will have a tough time realizing its the only one that doesn't get options. If we explicitly call out all types and map the others to none scala should let us know the match isn't exhaustive if new ones are added?

johnynek: Good call. Actually, we should never allow `case _`, it's just too unsafe.

ianoc: I imagine so, I imagine most of the other case _'s around producers at least are at risk of being invalid in future. I guess one exemption to that might be a sys.error

johnynek:
The storm tests are so damn flakey. We really need to solve #605

ianoc:
LGTM, merge when green

Issue #631, opened 2015-11-13
MemoryPlatform with unneeded .also can produce double writes


johnynek:
In a case that looks like:

```scala
val b = a.map(fn).sumByKey(store1)
val c = b.map(gn).sumByKey(store2)
b.also(c)
```
the also is not needed, since c already depends on b. It seems having it there is causing double counting.

Gabriel439:
Choosing `fn = gn = identity` and `store2 = NullStore` for simplicity, my understanding of this issue is that this pipeline:

```scala
(a.sumByKey(store1)).also(a.sumByKey(store1))
```

... will write out to `store1` twice.  Is that correct?

johnynek:
Yes, presumably that will show the bug as well (in the memory producer, not sure if it is a bug in other platforms).

Gabriel439:
The reason I asked is that I'm not sure whether or not I consider that behavior a bug.  If I read that I code I would expect it to write out to the store twice.

ianoc:
The contract at the producer layer is to be 'ensure' , i.e. ensure that other producer is included in the graph. you should be able to ensure the same bit of code as many times as you like only just ensuring its planned not that it runs more times

Gabriel439:
Oh, I see.  Alright, that sounds reasonable

Gabriel439:
Actually, I was thinking about this some more and I'm wondering how you can actually enforce that ensure contract given that you can transform producers.

Let me give a simple example:

```scala
(a.map(f).sumByKey(store)).also(a.map(g).sumByKey(store))

def f(x: T) = x
def g(x: T) = x
```

Even though `f` and `g` are the same function, how would summingbird know that those are two duplicate pipelines and avoid a double write?

ianoc:
those are two independent producers, regardless if the functions they are running are the same. Here you've explicitly created two sumByKey producers that are unique. even if they should be structurally equal.

the case that shouldn't break is

val p = a.map(g).sumByKey(store)

a.map(f).sumByKey(store).also(p).also(p).also(p)

Gabriel439:
So my concern then is that this breaks the law that:

```scala
p.map(identity) = p
```

... since attaching any function at all changes whether or not the producer is treated as a new independent producer.

johnynek:
So, one other thing: each store can only appear once in the graph as a sumByKey target.

[...]

PR #638, opened 2015-11-22
Add a test for redundant also in Memory platform

johnynek:
I tried to reproduce #631 but could not.

I wonder if this was actually user error, and we didn't notice.

I guess we should add this test, but it already passes.

Fixed a warning of a missing case match in MemoryPlatform, but we're not testing flatMapValues, obviously, since it has never been triggered.

### DIFF
--- a/summingbird-core-test/src/test/scala/com/twitter/summingbird/memory/MemoryLaws.scala
+++ b/summingbird-core-test/src/test/scala/com/twitter/summingbird/memory/MemoryLaws.scala
@@ -202,6 +202,18 @@ class MemoryLaws extends WordSpec {
     "lookupCollect w/ Int, Int" in { assert(lookupCollectChecker[Int, Int] == true) }
 
     "counters w/ Int, Int, Int" in { assert(counterChecker[Int, Int, Int] == true) }
+
+    "needless also shouldn't cause a problem" in {
+      val source = Memory.toSource(0 to 100)
+      val store1 = MutableMap.empty[Int, Int]
+      val store2 = MutableMap.empty[Int, Int]
+      val comp = source.map(v => (v % 3, v)).sumByKey(store1)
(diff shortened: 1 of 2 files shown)


summingbird-core-test/src/test/scala/com/twitter/summingbird/memory/MemoryLaws.scala
```diff
@@ -202,6 +202,18 @@ class MemoryLaws extends WordSpec {
     "lookupCollect w/ Int, Int" in { assert(lookupCollectChecker[Int, Int] == true) }
 
     "counters w/ Int, Int, Int" in { assert(counterChecker[Int, Int, Int] == true) }
+
+    "needless also shouldn't cause a problem" in {
+      val source = Memory.toSource(0 to 100)
+      val store1 = MutableMap.empty[Int, Int]
+      val store2 = MutableMap.empty[Int, Int]
+      val comp = source.map(v => (v % 3, v)).sumByKey(store1)
+      val prod = comp.also(comp.mapValues(_._2).write(new BufferFunc).sumByKey(store2))
```
ianoc: if you do the also a second time will it do anything funky?

ianoc:
LGTM, I had a quick question, but not a blocker, won't merge to make sure you see it. But feel free to merge.

summingbird-core-test/src/test/scala/com/twitter/summingbird/memory/MemoryLaws.scala
johnynek: I tried a bunch of stuff (which is why it is not trivial example up there). I started with `comp.also(comp)` and expanded out. I just ran it with an extra `.also(comp)` and that passed for me as well.

I think there is a chance he had a service backed by the same map somewhere in the job, which was the real issue (but I'm just guessing).

johnynek:
I'll merge because it is better to have a test, even though it does not seem to expose any issue.

PR #641, opened 2015-11-23
Remove catch Throwable, replace with NonFatal

johnynek:
It's possible this is hiding OOM errors in some cases. We really shouldn't catch OOM or linkage errors.


benpence:
LGTM

PR #642, opened 2015-11-23
Add logging to ConcurrentMemoryLaw tests

johnynek:
This is towards a fix on #604 

I can't get the test to fail locally, so I'm putting this in to see if it is the write or the store that is causing it to fail.


johnynek:
I ran the tests multiple times and never triggered the flakiness here. Also, since the value was a map, and there is the well known sparseness vs zero issue with naive map equality which I fixed here, we have a credible reason to believe this fixes #604.

I recommend we merge and reopen if we see the concurrent memory test flake again.

Issue #643, opened 2015-11-30
summingbird-example is broken


johnynek:
Looks like we are transitively pulling a later (3.4.x) branch of zookeeper, but storm 0.9.0-wip15 needs 3.3, so it fails at runtime.

It may work to exclude zookeeper from where we are getting it (maybe finagle via storehaus memcache or something).

ianoc:
Its odd we ran into this and our tests passed, it starts up a local ZK to run storm in local mode for the tests....

johnynek:
I assume that something in the example is exercising in a way that we aren't in the storm tests (for instance, we use storehaus-memcache in the example).

ianoc:
Yeah, though we have a storehaus dependency everywhere, and it all depends on util. (Though maybe not finagle i suppose). Just a little concerned about the utility of the tests. But i guess transitive dependency hell is what it is...

PR #644, opened 2015-11-30
Make summingbird-storm compile with storm 0.10.0

johnynek:
Did this to address #643. Had to make a few changes due to slf4j changes that I guess get pulled in transitively (man, transitive deps are kind of a broken idea, I'm afraid, who knows what other versions get bumped with this change?).

The real issue is that storm 0.9.0-wip15 was using an old zookeeper that is incompatible with recent finagle, which is also used in the examples. The solution to #643 then is to upgrade storm to get on a compatible zookeeper.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -35,7 +35,10 @@ object SummingbirdBuild extends Build {
   val tormentaVersion = "0.11.0"
   val junitVersion = "4.11"
   val log4jVersion = "1.2.16"
-  val stormVersion = "0.9.0-wip15"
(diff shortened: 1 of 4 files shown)


ianoc:
Shame so many seemingly unrelated changes are required, that said they look fine to me. I don't believe log4j logging is call by name though so moving to interpolation might make the logging more expensive i guess. But its all log.info we expect people to not be on performance critical paths anyway.

LGTM, merge when green

johnynek:
@ianoc I think all the logging is out of the critical path (just startup/plan logging).

The change I made in FlatMapBoltProvider was just to kill a warning while I was at things.

Issue #646, opened 2015-12-04
infinite loop in Dependants (or DagOptimizer?)


johnynek:
User hit this in ConcurrentMemoryPlatform (which uses the DagOptimizer):
```
scala.collection.immutable.List.flatMap(List.scala:327)
                    at com.twitter.summingbird.Dependants.dependantsAfterMerge(Dependants.scala:50)
                    at com.twitter.summingbird.Dependants$$anonfun$dependantsAfterMerge$2.apply(Dependants.scala:52)
                    at com.twitter.summingbird.Dependants$$anonfun$dependantsAfterMerge$2.apply(Dependants.scala:50)
                    at scala.collection.immutable.List.flatMap(List.scala:327)
                    at com.twitter.summingbird.Dependants.dependantsAfterMerge(Dependants.scala:50)
                    at com.twitter.summingbird.Dependants$$anonfun$dependantsAfterMerge$2.apply(Dependants.scala:52)
                    at com.twitter.summingbird.Dependants$$anonfun$dependantsAfterMerge$2.apply(Dependants.scala:50)
                    at scala.collection.immutable.List.flatMap(List.scala:327)
                    at com.twitter.summingbird.Dependants.dependantsAfterMerge(Dependants.scala:50)
                    at com.twitter.summingbird.Dependants$$anonfun$dependantsAfterMerge$2.apply(Dependants.scala:52)
                    at com.twitter.summingbird.Dependants$$anonfun$dependantsAfterMerge$2.apply(Dependants.scala:50)
                    at scala.collection.immutable.List.flatMap(List.scala:327)
                    at com.twitter.summingbird.Dependants.dependantsAfterMerge(Dependants.scala:50)
                    at com.twitter.summingbird.Dependants$$anonfun$dependantsAfterMerge$2.apply(Dependants.scala:52)
                    at com.twitter.summingbird.Dependants$$anonfun$dependantsAfterMerge$2.apply(Dependants.scala:50)
                    at scala.collection.immutable.List.flatMap(List.scala:327)
                    at com.twitter.summingbird.Dependants.dependantsAfterMerge(Dependants.scala:50)
                    at com.twitter.summingbird.Dependants$$anonfun$dependantsAfterMerge$2.apply(Dependants.scala:52)
                    at com.twitter.summingbird.Dependants$$anonfun$dependantsAfterMerge$2.apply(Dependants.scala:50)
                    at scala.collection.immutable.List.flatMap(List.scala:327)
[...]

Issue #645, opened 2015-12-04
Memory platform misbehaves for with use of also


pankajroark:
Test that fails:
```scala
class SbAlsoProducerTest extends TestBase {

  case class Key(i: Int)
  case class Value(v: Int)

  test("Two events for same entity combo should calculate correctly") {
    val platform = new Memory
    val source = Memory.toSource\[Value\](Seq(1, 2).map(Value(_)))
    val store: Memory#Store[Key, Value] = mutable.Map.empty[Key, Value]
    val sink: Memory#Sink[Value] = _ => ()

    implicit val sgrp = new Semigroup[Value] {
      override def plus(l: Value, r: Value): Value = Value(l.v + r.v)
    }

    val summed = source
      .map { v => (Key(v.v), v)}
      .sumByKey(store)
      .map { case (_, (existingEventOpt, currentEvent)) =>
        println(s"existing event $existingEventOpt")
        println(s"current event $currentEvent")
        existingEventOpt.map { existingEvent =>
          sgrp.plus(existingEvent, currentEvent)
        }.getOrElse(currentEvent)
      }

    val write1 = summed.write(sink)
    val write2 = summed.write(sink)
    val job = write1.also(write2)

    platform.run(platform.plan(job))
    assert(1 == store(Key(1)))
    assert(2 == store(Key(2)))
  }
}
```
I see that the map with print statements get called 4 times instead of two.

pankajroark:
Related issue #610

johnynek:
related to #631

johnynek:
Thanks for finding a case that fails. I'll look into it.

PR #647, opened 2015-12-10
Fix a bug in memory platforms with redundant also

### LINKED ISSUE #631: MemoryPlatform with unneeded .also can produce double writes

johnynek:
This adds the tests in #645, verifies that bug manifests, then I fixed the bug. Seems to fix #631 for real. /cc @pankajroark

### DIFF
--- a/summingbird-core/src/main/scala/com/twitter/summingbird/Dependants.scala
+++ b/summingbird-core/src/main/scala/com/twitter/summingbird/Dependants.scala
@@ -42,14 +42,16 @@ case class Dependants[P <: Platform[P]](tail: Producer[P, Any]) extends Dependan
   }
 
   /**
-   * Return the first dependants of this node AFTER merges. This will include
-   * no MergedProducer nodes, but all parents of the returned nodes will be Merged
-   * or the argument.
+   * Return the first dependants of this node AFTER merges and skipping Alsos.
+   * This will include no MergedProducer or AlsoProducer nodes, but all dependencies
+   * of the returned nodes will be Merged, Alsos or the argument.
+   * The purpose of this method is for source-down platforms
+   * that plan from sources to all their dependants recursively.
    */
   def dependantsAfterMerge(inp: Producer[P, Any]): List[Producer[P, Any]] =
     dependantsOf(inp).getOrElse(Nil).flatMap {
       case m @ MergedProducer(_, _) => dependantsAfterMerge(m)
-      case AlsoProducer(_, r) => dependantsAfterMerge(r)
+      case a @ AlsoProducer(_, _) => dependantsAfterMerge(a)
       case prod => List(prod)
     }
 
--- a/summingbird-core/src/main/scala/com/twitter/summingbird/memory/Memory.scala
+++ b/summingbird-core/src/main/scala/com/twitter/summingbird/memory/Memory.scala
@@ -16,8 +16,8 @@
 
 package com.twitter.summingbird.memory
 
-import com.twitter.algebird.Monoid
 import com.twitter.summingbird._
+import com.twitter.summingbird.graph.HMap
 import com.twitter.summingbird.option.JobId
 import com.twitter.summingbird.planner.DagOptimizer
 import collection.mutable.{ Map => MutableMap }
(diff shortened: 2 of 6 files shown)


johnynek:
It also fixes #646

summingbird-core/src/main/scala/com/twitter/summingbird/Dependants.scala
```diff
@@ -42,14 +42,16 @@ case class Dependants[P <: Platform[P]](tail: Producer[P, Any]) extends Dependan
...
+   * This will include no MergedProducer or AlsoProducer nodes, but all dependencies
+   * of the returned nodes will be Merged, Alsos or the argument.
+   * The purpose of this method is for source-down platforms
+   * that plan from sources to all their dependants recursively.
    */
   def dependantsAfterMerge(inp: Producer[P, Any]): List[Producer[P, Any]] =
     dependantsOf(inp).getOrElse(Nil).flatMap {
       case m @ MergedProducer(_, _) => dependantsAfterMerge(m)
-      case AlsoProducer(_, r) => dependantsAfterMerge(r)
+      case a @ AlsoProducer(_, _) => dependantsAfterMerge(a)
```
johnynek: here is the bug that hits `ConcurrentMemory`.

summingbird-core/src/main/scala/com/twitter/summingbird/memory/Memory.scala
```diff
@@ -82,11 +82,11 @@ class Memory(implicit jobID: JobId = JobId("default.memory.jobId")) extends Plat
           case AlsoProducer(l, r) =>
             //Plan the first one, but ignore it
             val (left, leftM) = toStream(l, jamfs)
+            val (right, rightM) = toStream(r, leftM)
```
johnynek: here is the bug in Memory producer. You should make the `lforcedEmpty` lazy or be sure to plan the right before you evaluate it. Here I do both since it seemed more correct to plan as much as possible before causing side effects, but also that we don't want side-effects at all until the job is run.

The underlying issue is that `Stream` is not entirely lazy. The head is eager. So, if you do `.filter(_ => false)` there will be no head but it evaluates as much as possible until it finds a head, which means it evaluates the whole stream. By evaluating the steam it hits the side effects in the map functions on `sumByKey` and `write`.

summingbird-core/src/main/scala/com/twitter/summingbird/Dependants.scala
ianoc: The bug here is that the dependantsOf(p) != r in this case?

johnynek: yes, I was looking up the graph, then dependantsOf brought me back down, causing an infinite loop. Wrong direction.

summingbird-core/src/main/scala/com/twitter/summingbird/Dependants.scala
```diff
@@ -42,14 +42,16 @@ case class Dependants[P <: Platform[P]](tail: Producer[P, Any]) extends Dependan
   }
 
   /**
-   * Return the first dependants of this node AFTER merges. This will include
-   * no MergedProducer nodes, but all parents of the returned nodes will be Merged
-   * or the argument.
+   * Return the first dependants of this node AFTER merges and skipping Alsos.
+   * This will include no MergedProducer or AlsoProducer nodes, but all dependencies
```
ianoc: s/all dependencies/all dependants ?

ianoc:
LGTM, maybe minor comment update if the revision makes sense.

summingbird-core/src/main/scala/com/twitter/summingbird/Dependants.scala
johnynek: no, I think that is right. It is saying if you look up (towards dependencies) for all the returned nodes, they will either be Merged, Also or inp. I could add a test to verify that. In fact I should (but the test I added was related).

ianoc: Yeah, seems right

PR #663, opened 2016-06-06
Fix outstanding futures clear bottleneck

pankajroark:
When maxWaitingFutures is high, AsyncBase can end up spending a lot of compute just clearing finished futures. The cost of clearing the finished futures is proportional to total number of outstanding futures.  If the number of finished futures is very low compared to total outstanding then we can end up paying this price for nothing much. And the same pattern can repeat on every execute. I found this happening in one of the jobs. In this change we keep track of number of pending futures and use that info to make sure that we clear the finished futures only when they are a good proportion of total number of outstanding futures so that we get good return on investment.

### DIFF
--- a/summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala
+++ b/summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala
@@ -16,12 +16,25 @@ limitations under the License.
 
 package com.twitter.summingbird.online.executor
 
+import java.util.concurrent.atomic.AtomicInteger
+
 import com.twitter.summingbird.online.Queue
 import com.twitter.summingbird.online.option.{ MaxWaitingFutures, MaxFutureWaitTime, MaxEmitPerExecute }
 import com.twitter.util.{ Await, Future }
 import scala.util.{ Try, Success, Failure }
 import java.util.concurrent.TimeoutException
 import org.slf4j.{ LoggerFactory, Logger }
+
+object AsyncBase {
+  /**
+   * Ratio of total number of outstanding futures to the portion that is finished
+   * ,at which finished futures are cleared.
+   * Clearing finished futures costs proportional to total number of outstanding
+   * futures, so we want to make sure we only clear when sufficient portion is
+   * finished.
+   */
+  val OutstandingFuturesDequeueRatio = 2
+}
 
 abstract class AsyncBase[I, O, S, D, RC](maxWaitingFutures: MaxWaitingFutures, maxWaitingTime: MaxFutureWaitTime, maxEmitPerExec: MaxEmitPerExecute) extends Serializable with OperationContainer[I, O, S, D, RC] {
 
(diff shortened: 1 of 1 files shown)


summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala
```diff
@@ -80,20 +94,32 @@ abstract class AsyncBase[I, O, S, D, RC](maxWaitingFutures: MaxWaitingFutures, m
   private def addOutstandingFuture(fut: Future[Unit]): Boolean =
     if (!fut.isDefined) {
       outstandingFutures.put(fut)
+      numPendingOutstandingFutures.addAndGet(1)
+      fut.respond { _ => numPendingOutstandingFutures.addAndGet(-1) }
```
johnynek: can we use `.ensure(numPendingOutstandingFutures.decrementAndGet())`?

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala
```diff
@@ -80,20 +94,32 @@ abstract class AsyncBase[I, O, S, D, RC](maxWaitingFutures: MaxWaitingFutures, m
   private def addOutstandingFuture(fut: Future[Unit]): Boolean =
     if (!fut.isDefined) {
       outstandingFutures.put(fut)
+      numPendingOutstandingFutures.addAndGet(1)
```
johnynek: can we use `numPendingOutstandingFutures.incrementAndGet()`

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala
```diff
@@ -16,13 +16,26 @@ limitations under the License.
...
 
+object AsyncBase {
+  /**
+   * Ratio of total number of outstanding futures to the portion that is finished
+   * ,at which finished futures are cleared.
+   * Clearing finished futures costs proportional to total number of outstanding
+   * futures, so we want to make sure we only clear when sufficient portion is
+   * finished.
+   */
+  val OutstandingFuturesDequeueRatio = 2
```
johnynek, line 36: should this be a tunable option with a default of 2?

johnynek:
couple of small comments.

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala
johnynek, line 36: maybe not for now. We can revisit if this does not fix the issue. Less knobs are better.

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala
pankajroark: Good idea.

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala
pankajroark, line 36: I agree, making this a knob seemed overkill to me as well.

pankajroark:
@johnynek I incorporated comments, could you take a look again? Thanks

PR #665, opened 2016-06-14
Incorporate latest storehaus release

pankajroark:
Corresponding changes to changes in FutureCollector type signature.


rubanm:
@pankajroark  Looks like some tests are failing on travis
There has been no public 0.10.0 release (only RCs) so we may not strictly need to change to 0.11.0-SNAPSHOT, but it doesn't hurt I guess.

pankajroark:
Fixed tests. Storehaus depends on latest versions of finagle and util, updated those. Example project was failing due to some excludes. I've removed those excludes. Not sure what the excludes were meant for but don't seem to be needed now.

rubanm:
👍

johnynek:
looks like this was added here:

https://github.com/twitter/summingbird/commit/4a80201c9b30d2dd9d034e898df8f0fb6b4f8ebc

@egonina do you remember why it was added?

PR #667, opened 2016-06-20
Fix root cause of duplicate summer execution in Memory platform

pankajroark:
AlsoProducers. The root cause has actually to do with the summer. In
toStream method we keep a map of Producer to stream to make sure we don't run
parts of the graph again and again. The map uses value equality. The
problem is that summer includes store which is mutable. If a summer is
referenced by two also producers and if by the time summer is revisited
the store has changed then the key would be different from the original
summer and won't be found in the map, leading to the summer being
planned again. d3b5808 seems to fix the issue and passes a unit test
because that test has only one AlsoProducer, where planning left and
right before forcing left means that store doesn't change in between.
But it likely wouldn't fix the case where there are multiple AlsoProducers at
different levels. Even if it did fix those cases this is an indirect
solution and it's better to avoid taking any chances and fix the root of
the issue.

This is a case where reference equality fits the bill. I don't know of
an immutable implementation of a map that uses reference equality so
I'm using java.util.IdentityHashMap, if there is one I would love to use
that instead.

### DIFF
--- a/summingbird-core/src/main/scala/com/twitter/summingbird/memory/Memory.scala
+++ b/summingbird-core/src/main/scala/com/twitter/summingbird/memory/Memory.scala
@@ -46,7 +46,7 @@ class Memory(implicit jobID: JobId = JobId("default.memory.jobId")) extends Plat
   def counter(group: Group, name: Name): Option[Long] =
     MemoryStatProvider.getCountersForJob(jobID).flatMap { _.get(group.getString + "/" + name.getString).map { _.get } }
 
-  def toStream[T, K, V](outerProducer: Prod[T], jamfs: JamfMap): (Stream[T], JamfMap) =
+  def toStream[T](outerProducer: Prod[T], jamfs: JamfMap): (Stream[T], JamfMap) =
     Option(jamfs.get(outerProducer)) match {
       case Some(s) => (s.asInstanceOf[Stream[T]], jamfs)
       case None =>
--- a/summingbird-core/src/main/scala/com/twitter/summingbird/memory/Memory.scala
+++ b/summingbird-core/src/main/scala/com/twitter/summingbird/memory/Memory.scala
@@ -16,8 +16,10 @@
 
 package com.twitter.summingbird.memory
 
+import java.util
+
+import com.twitter.algebird.Monoid
 import com.twitter.summingbird._
-import com.twitter.summingbird.graph.HMap
 import com.twitter.summingbird.option.JobId
 import com.twitter.summingbird.planner.DagOptimizer
 import collection.mutable.{ Map => MutableMap }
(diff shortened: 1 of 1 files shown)


summingbird-core/src/main/scala/com/twitter/summingbird/memory/Memory.scala
```diff
@@ -39,82 +41,85 @@ class Memory(implicit jobID: JobId = JobId("default.memory.jobId")) extends Plat
...
-                store.update(k, newV)
-                (k, (oldV, deltaV))
-            }
-            (summed, m)
-        }
-        // scala can't infer that s is the right type through case statements above
-        val st = s.asInstanceOf[Stream[T]]
-        (st, m + (outerProducer -> st))
+  def toStream[T, K, V](outerProducer: Prod[T], jamfs: JamfMap): (Stream[T], JamfMap) = {
+    val stream = jamfs.get(outerProducer)
```
johnynek: could we do: `Option(jamfs.get(outerProducer)) match {` so that the diff can be smaller? This is harder to read.

summingbird-core/src/main/scala/com/twitter/summingbird/memory/Memory.scala
```diff
@@ -39,14 +41,14 @@ class Memory(implicit jobID: JobId = JobId("default.memory.jobId")) extends Plat
...
-  private type JamfMap = HMap[Prod, Stream]
+  private type JamfMap = util.IdentityHashMap[Prod[_], Stream[_]]
 
   def counter(group: Group, name: Name): Option[Long] =
     MemoryStatProvider.getCountersForJob(jobID).flatMap { _.get(group.getString + "/" + name.getString).map { _.get } }
 
-  private def toStream[T](outerProducer: Prod[T], jamfs: JamfMap): (Stream[T], JamfMap) =
-    jamfs.get(outerProducer) match {
-      case Some(s) => (s, jamfs)
+  def toStream[T, K, V](outerProducer: Prod[T], jamfs: JamfMap): (Stream[T], JamfMap) =
```
jnievelt: K, V don't seem to be used?

summingbird-core/src/main/scala/com/twitter/summingbird/memory/Memory.scala
```diff
@@ -99,21 +101,22 @@ class Memory(implicit jobID: JobId = JobId("default.memory.jobId")) extends Plat
             }
             (joined, m)
 
-          case Summer(producer, store, semigroup) =>
+          case Summer(producer, store, monoid) =>
             val (s, m) = toStream(producer, jamfs)
             val summed = s.map {
-              case (k, deltaV) =>
+              case pair @ (k, deltaV) =>
```
jnievelt: pair not used?  Really this block could be left unchanged

pankajroark:
Some context for the review. This change is basically a revert of d3b5808 followed by use of IdentityHashMap instead of Map for JamfMap. btw, I don't know what Jamf stands for and curios find out.

summingbird-core/src/main/scala/com/twitter/summingbird/memory/Memory.scala
pankajroark: Good catch, fixed.

summingbird-core/src/main/scala/com/twitter/summingbird/memory/Memory.scala
```diff
@@ -39,14 +41,14 @@ class Memory(implicit jobID: JobId = JobId("default.memory.jobId")) extends Plat
   type Plan[T] = Stream[T]
 
   private type Prod[T] = Producer[Memory, T]
-  private type JamfMap = HMap[Prod, Stream]
+  private type JamfMap = util.IdentityHashMap[Prod[_], Stream[_]]
```
johnynek: this is mutable now, and before it was immutable. Does this not cause any problems? Can you add some comments explaining how it is safe? Also, why thread it through if you are mutating it? There is no need to return `JamfMap` if you are just returning the input.

I would prefer not introduce a mutable map, and instead add:

``` scala
class Identity[T](val unwrap: T) {
  override def equals(that: Any) = that match {
    case i: Identity[_] => unwrap eq i.unwrap
    case _ => false
  }
  override def hashCode = System.identityHashCode(unwrap)
}
private type Prod[T] = Identity[Producer[Memory, T]]
```

then wrap keys with `new Identity(key)`.

I really hate to give up the reasoning we get with immutable types, and we should only do so for big performance wins, but we don't care about performance here (planning is fast, and only happens at submit).

summingbird-core/src/main/scala/com/twitter/summingbird/memory/Memory.scala
```diff
@@ -99,21 +101,22 @@ class Memory(implicit jobID: JobId = JobId("default.memory.jobId")) extends Plat
...
+                val newV = oldV.map {
+                  monoid.plus(_, deltaV)
+                }
                   .getOrElse(deltaV)
                 store.update(k, newV)
                 (k, (oldV, deltaV))
             }
             (summed, m)
         }
-        // scala can't infer that s is the right type through case statements above
```
johnynek, line 114: this comment is still true (note you have a cast).

johnynek:
`Jamf` was an old inside joke between @singhala @sritchie and I based on Jules Winnfield's wallet, and changing the word "Bad" to "Jive Ass".

johnynek:
https://www.youtube.com/watch?v=NUuwd8Z0l_4

summingbird-core/src/main/scala/com/twitter/summingbird/memory/Memory.scala
pankajroark: This is a good idea. I would really love to keep everything immutable too. Let me give this a shot.

pankajroark:
He he, I love pulp fiction dialogs.

I've updated the review to use immutable map with reference equality.

summingbird-core/src/main/scala/com/twitter/summingbird/memory/Memory.scala
```diff
@@ -82,11 +116,11 @@ class Memory(implicit jobID: JobId = JobId("default.memory.jobId")) extends Plat
           case AlsoProducer(l, r) =>
             //Plan the first one, but ignore it
             val (left, leftM) = toStream(l, jamfs)
-            val (right, rightM) = toStream(r, leftM)
             // We need to force all of left to make sure any
             // side effects in write happen
-            lazy val lforcedEmpty = left.filter(_ => false)
-            (right.append(lforcedEmpty), rightM)
+            val lforcedEmpty = left.filter(_ => false)
+            val (right, rightM) = toStream(r, leftM)
```
johnynek: why do we have to revert this change? Calling `toStream(r, leftM)` first seems more correct to me, as does using the lazy concatenation (rather than the strict that we have reverted to).

pankajroark: It breaks a bunch of our tests which are relying on the current execution order. Since we don't need to change the ordering to fix the duplication issue it will be great if we can preserve the same execution order.

johnynek: As we discussed, can you just copy this file into the repo rather than hardcode a bug just because a few tests assumed something false?

I fear the "keeping bugs for a few tests" is not scalable. I doubt you would be happy if you hit a bug that I want us to keep because my tests assumed the buggy behavior.

johnynek: just to link to the prior discussion:
https://github.com/twitter/summingbird/pull/647/files#r47308385

#647 may indeed be fixed more than one way, but I still don't like that planning forces the left hand side.

Could you get the tests to pass if you did something like:

``` scala
def merge[A, B](a: Stream[A], b: Stream[B]): Stream[Either[A, B]] = {
  val itA = a.iterator
  val itB = b.iterator
  var left = itA.hasNext
    // try to alternate as much as possible
  def next: Option[Either[A, B]] =
    if (!(itA.hasNext || itB.hasNext)) None
    else if (left) {
      left = false
      if (itA.hasNext) Some(Left(itA.next)) else next
    } else {
      left = true
      if (itB.hasNext) Some(Right(itB.next)) else next
    }
  Stream.continually(next).takeWhile(_.isDefined).map(_.get)
}
```

Then we can return `merge(left, right).collect { case Right(r) => r }`. as the result of the `Also`. In this way, we are closer to modeling a real streaming system.

If the contract we are presenting here is that the left will always be consumed at planning, I just don't like that. It leads to a side-effect from planning which was never our intent. I think the test needs to not assume that.

I think we need to make a clear case as to why the behavior is a certain way to which is not "some broken test at Twitter will fail if we change this". If that is the only reason, you should have an internal fork of the platform.

pankajroark: One of the broken tests actually got fixed by this cl. Other teams have agreed to work on fixing their tests or comment out the broken test for now. I've updated the cl to restore the lazy evaluation of left producer. Please take a look.

summingbird-core/src/main/scala/com/twitter/summingbird/memory/Memory.scala
```diff
@@ -16,6 +16,9 @@
 
 package com.twitter.summingbird.memory
 
+import java.util
```
johnynek: we aren't using this anymore are we?

summingbird-core/src/main/scala/com/twitter/summingbird/memory/Memory.scala
```diff
@@ -16,6 +16,9 @@
...
+
+import com.twitter.algebird.Monoid
```
johnynek: is this needed?

johnynek:
Okay, given that this was always an example and toy platform, if it is really important for twitter for it to have the (unrealistic) semantics that in an also, the left is fully evaluated before the right, I'm okay with that (since it didn't cause an issue for anyone that we know of yet).

If the tests can be fixed to not assume that, even better.

I'll leave it up to @pankajroark 

👍

summingbird-core/src/main/scala/com/twitter/summingbird/memory/Memory.scala
pankajroark: My bad, have been relying on intellij to eliminate these. intellij gave up at some point and I didn't relize. Removed now.

pankajroark:
In this case we've decided to remove/fix the tests that are failing due to wrong assumptions. So we're going ahead and keeping the change to make execution of left producer lazy.

sritchie:
`Jamf` is exactly what @johnynek says; it's also a reference to Dr. Laszlo Jamf, from Gravity's Rainbow: https://en.wikipedia.org/wiki/Gravity%27s_Rainbow

pankajroark:
@sritchie Never heard of Gravity's rainbow, I've got to read it now :)

PR #668, opened 2016-06-27
Externalizing optionMap collapse and metrics registration for StormPlatform

jnievelt:
For mapping bolts, we generally use `FlatMapOperation` which wraps the mapping function in an `Externalizer`.  We don't currently do this for functions that go directly into the spout, however.

This change wraps both the `optionMap` function and the `registerAllMetrics` function in `Externalizer`s before giving them to the spout.

This allows, for example, using `Source#collect` with Storm (i.e., pushing a partial function into the spout).

### DIFF
--- a/summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormPlatform.scala
+++ b/summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormPlatform.scala
@@ -172,7 +172,9 @@ abstract class Storm(options: Map[String, Options], transformConfig: Summingbird
     val tormentaSpout = node.members.reverse.foldLeft(spout.asInstanceOf[Spout[(Timestamp, Any)]]) { (spout, p) =>
       p match {
         case Source(_) => spout // The source is still in the members list so drop it
-        case OptionMappedProducer(_, op) => spout.flatMap { case (time, t) => op.apply(t).map { x => (time, x) } }
+        case OptionMappedProducer(_, op) =>
+          val boxed = Externalizer(op)
+          spout.flatMap { case (time, t) => boxed.get(t).map { x => (time, x) } }
         case NamedProducer(_, _) => spout
         case IdentityKeyedProducer(_) => spout
         case AlsoProducer(_, _) => spout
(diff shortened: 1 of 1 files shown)


summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormPlatform.scala
```diff
@@ -184,19 +187,17 @@ abstract class Storm(options: Map[String, Options], transformConfig: Summingbird
...
+        x: StormMetric[IMetric] =>
+          context.registerMetric(x.name, x.metric, x.interval.inSeconds)
       }
-    }
-    val stormSpout = tormentaSpout.openHook(registerAllMetrics).getSpout
+      // Register summingbird counter metrics.
+      StormStatProvider.registerMetrics(jobID, context, countersForSpout)
+      SummingbirdRuntimeStats.addPlatformStatProvider(StormStatProvider)
+    })
+    val stormSpout = tormentaSpout.openHook(registerAllMetrics.get).getSpout
```
johnynek, line 199: how does this help if we are just calling `.get`? Are you sure this is needed? Would marking it `@transient` also work instead (and be simpler)?

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormPlatform.scala
```diff
@@ -172,7 +172,10 @@ abstract class Storm(options: Map[String, Options], transformConfig: Summingbird
     val tormentaSpout = node.members.reverse.foldLeft(spout.asInstanceOf[Spout[(Timestamp, Any)]]) { (spout, p) =>
       p match {
         case Source(_) => spout // The source is still in the members list so drop it
-        case OptionMappedProducer(_, op) => spout.flatMap { case (time, t) => op.apply(t).map { x => (time, x) } }
+        case OptionMappedProducer(_, op) => {
```
johnynek: we usually don't use `{ }` in case blocks.

johnynek:
looks good to me other than some small comments.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormPlatform.scala
jnievelt, line 199: The openHook param is passed by name, so get is actually called on the other side (where the spout is opened).

I'm not sure about how to make this @transient (AFAIK that applies to class members, while this is function-local).

TBH I'm not sure why there was suddenly an issue serializing this function.  I can poke more into it.

johnynek, line 199: okay, if it is call-by-name, I guess this is fine.

jnievelt, line 199: Was looking around this today; it's possible this doesn't cause an issue right now because the default `Spout` implementation for openHook is to ignore it:

https://github.com/twitter/tormenta/blob/v0.11.1/tormenta-core/src/main/scala/com/twitter/tormenta/spout/Spout.scala#L49

I ran into issues with this during some testing I was doing, probably because I was using `BaseSpout` which actually keeps the hook:

https://github.com/twitter/tormenta/blob/v0.11.1/tormenta-core/src/main/scala/com/twitter/tormenta/spout/BaseSpout.scala#L32

Maybe `BaseSpout` could do the externalizing, but it seems like all spout implementations (that do open hooks) would need to as well.

johnynek:
👍

johnynek:
@jnievelt what is the story here? Should we not merge?

jnievelt:
I'll squash this and merge today.

PR #670, opened 2016-06-28
Pg/force futures better

pankajroark:
Sending out early for review, will add tests soon.


summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala
```diff
@@ -34,6 +35,30 @@ object AsyncBase {
...
+   */
+  def waitN[A](fs: Queue[Future[A]], n: Int): Future[Unit] = {
+    require(n >= 0)
+    val waitOnCount = Math.min(fs.size, n)
+    if (waitOnCount == 0) {
+      Future.Unit
+    } else {
+      val count = new AtomicInteger(waitOnCount)
+      val p = Promise[Unit]()
+      for (f <- fs) {
```
johnynek: can we use `fs.foreach` here?

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala
```diff
@@ -34,6 +35,30 @@ object AsyncBase {
...
    */
   val OutstandingFuturesDequeueRatio = 2
+
+  /**
+   * Wait for n futures to finish. Doesn't block, the returned future is satisfied
+   * once n futures have finished either successfully or unsuccessfully.
+   * If n is greater than number of futures in queue then we wait on all of them.
+   */
+  def waitN[A](fs: Queue[Future[A]], n: Int): Future[Unit] = {
+    require(n >= 0)
```
johnynek: we could just return `Future.Unit` when `n <= 0` similar to how `drop(n)` is identity when `n <= 0`.

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala
```diff
@@ -34,6 +35,30 @@ object AsyncBase {
...
+    val waitOnCount = Math.min(fs.size, n)
+    if (waitOnCount == 0) {
+      Future.Unit
+    } else {
+      val count = new AtomicInteger(waitOnCount)
+      val p = Promise[Unit]()
+      for (f <- fs) {
+        f ensure {
```
johnynek: can we use `f.ensure`? I prefer that we use `.` since without it sometimes you hit weird `;` inference problems which confuse newcomers.

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala
```diff
@@ -34,6 +35,30 @@ object AsyncBase {
...
+          if (count.decrementAndGet() == 0) {
+            p.setValue(())
```
johnynek, line 56: can we add a comment here, something like:

```
Note, since we are only decrementing, we can only cross the value 0 one time (unless we decrement > 2^(32) times).
```

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala
johnynek: this will consume all the items from the Queue. Is that the intent? Can we add that comment to the method call?

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala
```diff
@@ -34,6 +35,30 @@ object AsyncBase {
    * finished.
    */
   val OutstandingFuturesDequeueRatio = 2
+
+  /**
+   * Wait for n futures to finish. Doesn't block, the returned future is satisfied
+   * once n futures have finished either successfully or unsuccessfully.
+   * If n is greater than number of futures in queue then we wait on all of them.
+   */
+  def waitN[A](fs: Queue[Future[A]], n: Int): Future[Unit] = {
```
johnynek: maybe it is better to take `Iterable[Future[A]]` and use `Queue.toSeq` to mutate and take from the queue. Otherwise, it might not be clear we are mutating the Queue.

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala
```diff
@@ -107,13 +132,13 @@ abstract class AsyncBase[I, O, S, D, RC](maxWaitingFutures: MaxWaitingFutures, m
     if (pendingFuturesCount > maxWaitingFuturesCount) {
       // Too many futures waiting, let's clear.
       outstandingFutures.dequeueAll(_.isDefined)
-      val toForce = outstandingFutures.trimTo(maxWaitingFuturesCount).toIndexedSeq
-      if (toForce.nonEmpty) {
+      val toClear = outstandingFutures.size - maxWaitingFuturesCount
+      if (toClear > 0) {
         try {
-          Await.ready(Future.collect(toForce), maxWaitingTime.get)
+          Await.ready(AsyncBase.waitN(outstandingFutures, toClear), maxWaitingTime.get)
```
johnynek: I'm worried that this is emptying the queue, and even if we don't wait for all of them, and I'm worried we could have OOM issues with futures building up that are never satisfied.

johnynek: we could call `outstandingFutures.toSeq` then wait on `toClear` then requeue all the futures that are not defined.

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala
```diff
@@ -107,13 +132,13 @@ abstract class AsyncBase[I, O, S, D, RC](maxWaitingFutures: MaxWaitingFutures, m
...
         } catch {
           case te: TimeoutException =>
-            logger.error("forceExtra failed on %d Futures".format(toForce.size), te)
+            logger.error("forceExtra failed on %d Futures".format(toClear), te)
```
johnynek: can we use scala interpolation strings: `s"forceExtra failed on $toClear Futures"`

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala
pankajroark: I didn't realize that foreach on queue is removing dequing items from the queue. Removing items from the queue is expensive, I want to avoid doing that. Let me rethink this.

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala
pankajroark: My bad, didn't realize the foreach on queue was dequeing. I don't actually want to dequeue elements, it's expensive. I just want to iterate over them. Is that problematic on the queue?

johnynek: I'm not sure dequeueing is much more expensive than iterating. Iterating over a mutable collection is problematic. I think dequeue + requeue is clearer semantic.

With your previous check that we only do this when the number of finished futures is high, I don't think it will be expensive. I think we should try and see if perf is not good (maybe use yourkit) rather than possibly make it more confusing iterating through a mutable collection.

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala
```diff
@@ -135,10 +136,12 @@ abstract class AsyncBase[I, O, S, D, RC](maxWaitingFutures: MaxWaitingFutures, m
       val toClear = outstandingFutures.size - maxWaitingFuturesCount
       if (toClear > 0) {
         try {
-          Await.ready(AsyncBase.waitN(outstandingFutures, toClear), maxWaitingTime.get)
+          val dequeuedFutures = outstandingFutures.toSeq
+          Await.ready(AsyncBase.waitN(dequeuedFutures, toClear), maxWaitingTime.get)
+          outstandingFutures.putAll(dequeuedFutures.filter(_.isDefined))
```
johnynek: you want `outstandingFutures.putAll(dequeuedFutures.filterNot(_.isDefined))` right? You don't want to requeue if the future is complete, do you?

summingbird-online/src/test/scala/com/twitter/summingbird/online/AsyncBaseSpec.scala
```diff
@@ -0,0 +1,67 @@
...
+
+import com.twitter.summingbird.online.executor.AsyncBase
+import com.twitter.util._
+import org.scalatest.WordSpec
+import org.scalatest.concurrent.Eventually
+import org.scalatest.time.{ Millis, Span }
+
+class AsyncBaseSpec extends WordSpec with Eventually {
+
+  def onThread(f: => Unit): Thread = {
```
johnynek: you are not using this, are you?

summingbird-online/src/test/scala/com/twitter/summingbird/online/AsyncBaseSpec.scala
```diff
@@ -0,0 +1,67 @@
...
+    for (i <- 0 until Math.min(futuresCount, waitOn)) {
+      assert(t.unblocked === false)
+      valueToFill match {
+        case Return(v) =>
+          ps(i).setValue(v)
+        case Throw(e) =>
+          ps(i).setException(e)
+      }
+    }
+    eventually(timeout(Span(500, Millis)))(assert(t.unblocked === true))
```
johnynek: we might need a lot more time than this on travis. It is very slow and often timeouts like this become flakey.

pankajroark: Increased timeout.

summingbird-online/src/test/scala/com/twitter/summingbird/online/AsyncBaseSpec.scala
pankajroark: Removed.

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala
pankajroark: You're right. Fixed and added test.

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala
```diff
@@ -16,14 +16,15 @@ limitations under the License.
...
 import com.twitter.summingbird.online.Queue
-import com.twitter.summingbird.online.option.{ MaxWaitingFutures, MaxFutureWaitTime, MaxEmitPerExecute }
-import com.twitter.util.{ Await, Future }
-import scala.util.{ Try, Success, Failure }
-import java.util.concurrent.TimeoutException
-import org.slf4j.{ LoggerFactory, Logger }
+import com.twitter.summingbird.online.option.{ MaxEmitPerExecute, MaxFutureWaitTime, MaxWaitingFutures }
+import com.twitter.util._
+import org.slf4j.{ Logger, LoggerFactory }
+
```
johnynek, line 26: no need for the spaces. One big sorted block is fine.

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala
```diff
@@ -107,13 +136,15 @@ abstract class AsyncBase[I, O, S, D, RC](maxWaitingFutures: MaxWaitingFutures, m
     if (pendingFuturesCount > maxWaitingFuturesCount) {
       // Too many futures waiting, let's clear.
       outstandingFutures.dequeueAll(_.isDefined)
-      val toForce = outstandingFutures.trimTo(maxWaitingFuturesCount).toIndexedSeq
```
johnynek, line 110: one feature of the old algorithm is that it went in order, so one future could not wait too long. The new approach will allow a future to wait forever. Maybe that is okay, but it is a difference.

pankajroark, line 110: Could you explain a bit more on "one future could not wait too long".  The wait would have timed out with MaxFutureWaitTime, are you referring to that. In the new algo it can stay in the queue forever but that should be rare, futures would almost always have timeouts, if they don't then that's likely a bug.

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala
pankajroark, line 26: I was about to remove the spaces but then I saw that the whole project (summingbird-online) uses this format with spaces. Should I still remove spaces?

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala
johnynek, line 110: Yes, it would have timed out with MaxFutureWaitTime. Now, that is not true, so the users need to force timeouts on their futures, otherwise they could last quite long.

I think that will be okay, but I bet some folks will see weird corner cases. Be prepared.

That said, I'm optimistic this PR can really improve throughput quite a lot, so that's exciting.

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala
johnynek, line 26: yes. Old style.

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala
```diff
@@ -107,13 +136,15 @@ abstract class AsyncBase[I, O, S, D, RC](maxWaitingFutures: MaxWaitingFutures, m
     if (pendingFuturesCount > maxWaitingFuturesCount) {
       // Too many futures waiting, let's clear.
       outstandingFutures.dequeueAll(_.isDefined)
```
johnynek: maybe rather than `dequeueAll` we want to do:

``` scala
val pending = outstandingFutures.toSeq.filterNot(_.isDefined)
val toClear = pending.size - maxWaitingFuturesCount
if (toClear > 0) {
  // wait on N, put the ones still not defined at the end.
}
else {
  outstandingFutures.putAll(pending)
}
```

since `dequeueAll` internally does `toSeq` the above should be faster (and maybe a bit clearer).

pankajroark: I also noticed another issue which earlier implementation was accidentally avoiding. Since we'll now unblock immediately when we are below the limit we'll likely hit the limit pretty soon and then the high cost of iterating over all futures will come into picture. I think we should treat the limit as a high watermark and bring down the number of pending futures by x%, say 10%. This way we can amortize the cost of these operations. What do you think?

btw, your above suggestion looks great, I'm incorporating that.

johnynek: that may work well. It is hard to guess often. Is there a real job you can test with to see how it performs or a benchmark suite?

We could add a `dropWhile` method to queue and on every call do `dropWhile(_.isDefined)` to very cheaply prune. When a future on the front gets stuck, this could trigger the `waitN` behavior, but as long as things were humming along, `dropWhile` might be a simple way to keep costs low.

pankajroark: We might keep hitting the limit in terms of pending futures too. We get events in the bolt at a great rate, immediately after we clear the futures to get down to the limit we'll get another event and check again, by that time, or a few events later, we might again have more pending futures than the limit. Clearing defined futures will not help because the pending futures themselves may go over the limit. Having a lower watermark will avoid that.

I don't have a benchmark suite. I do have a job, profiling which I found the original issue. I guess it would be hard to see the effect of just the watermark change. But overall I do see improvement with all the fixes in this PR so far. I'm taking the yourkit profile to monitor how long the slave thread blocks waiting for the futures to clear. That period is certainly down, so the slave thread is consuming events at a greater rate.

pankajroark: I wrote a benchmark for a case with high store latency. I found that this change clearly increases throughput. Also the watermarks didn't make any difference. I've removed the watermark logic.

Here are the benchmark numbers before and after this change:
Before
SummerBenchmark.highLatencyStore  thrpt   10  **2077.829 ± 289.029  ops/s**

After
SummerBenchmark.highLatencyStore  thrpt   10  **3267.906 ± 139.766  ops/s**

I ran the benchmark a number of times and found the results to be consistent. 

Set up was:
Summer - SyncSummingQueue - buffersize 100, flush frequency 1 sec
MaxWaitingFutures - 1000
MaxFutureWaitTime - 5 sec
MaxEmitsPerExecute - Int.Max

The benchmark is a big change, I'm planning to create a separate PR for it after a couple weeks.
@johnynek PTAL

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala
```diff
@@ -106,15 +135,18 @@ abstract class AsyncBase[I, O, S, D, RC](maxWaitingFutures: MaxWaitingFutures, m
...
-      outstandingFutures.dequeueAll(_.isDefined)
-      val toForce = outstandingFutures.trimTo(maxWaitingFuturesCount).toIndexedSeq
-      if (toForce.nonEmpty) {
+      val pending = outstandingFutures.toSeq.filterNot(_.isDefined)
+      val toClear = pending.size - maxWaitingFuturesCount
+      if (toClear > 0) {
         try {
-          Await.ready(Future.collect(toForce), maxWaitingTime.get)
+          Await.ready(AsyncBase.waitN(pending, toClear), maxWaitingTime.get)
+          outstandingFutures.putAll(pending.filterNot(_.isDefined))
```
johnynek: I'm worried that is the `TimeoutException` happens, then we never put any futures back. I think in that case we should put them all back? What do you think about?

``` scala
try {
   Await.ready(AsyncBase.waitN(pending, toClear), maxWaitingTime.get)
} catch {
  case te: TimeoutException =>
}
outstandingFutures.putAll(pending.filterNot(_.isDefined))
```

pankajroark: Makes sense. Updated.

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala
```diff
@@ -16,14 +16,15 @@ limitations under the License.
...
+import java.util.concurrent.TimeoutException
 import java.util.concurrent.atomic.AtomicInteger
 
 import com.twitter.summingbird.online.Queue
-import com.twitter.summingbird.online.option.{ MaxWaitingFutures, MaxFutureWaitTime, MaxEmitPerExecute }
-import com.twitter.util.{ Await, Future }
-import scala.util.{ Try, Success, Failure }
-import java.util.concurrent.TimeoutException
-import org.slf4j.{ LoggerFactory, Logger }
+import com.twitter.summingbird.online.option.{MaxEmitPerExecute, MaxFutureWaitTime, MaxWaitingFutures}
```
pankajroark, line 23: Running tests automatically restores the old formatting. Any tips on how to go about disabling that?

johnynek, line 23: we use scalariform to fix the formatting. The goal is not to fight it.

:)

Maybe we should move to scalafmt:
https://github.com/olafurpg/scalafmt

johnynek:
👍 

merge when green.

Nice work working through this and measuring the improvement. Can you possibly think about adding a benchmarking system to the build so we can re-run those things (and maybe even thing about how to do a travis check of performance).

pankajroark:
Yes, I ultimately plan to add benchmarks here. For now I'm working in our internal repo to refine/grow these benchmarks. We are woefully short of performance testing for online summingbird, want to fix that internally first and then contribute here.

Issue #671, opened 2016-06-29
Summingbird doesn't pick correct option when same setting defined on multiple nodes


pankajroark:
I was trying to apply different MaxWaitingFutures on the flatMapper and summer on a simple source -> flatMap -> sink topology. The setting from flatMapper is getting applied to summer. 

I was able to simulate this in a unit test.
This test fails when added to TopologyTests:
  "With same setting on multiple names we use the one for the node" in {
    val fmNodeName = "flatMapper"
    val smNodeName = "summer"
    val p = Storm.source(TraversableSpout(sample[List[Int]]))
      .flatMap(testFn).name(fmNodeName)
      .sumByKey(TestStore.createStore[Int, Int]()._2).name(smNodeName)

```
val opts = Map(fmNodeName -> Options().set(SummerParallelism(10)),
  smNodeName -> Options().set(SummerParallelism(20)))
val storm = Storm.local(opts)
val stormTopo = storm.plan(p).topology
val bolts = stormTopo.get_bolts

// Tail should use parallelism specified for the summer node
assert(bolts("Tail").get_common.get_parallelism_hint == 20)
```

  }

flatMap node's setting for SummerParallelism overrides that of summer.

pankajroark:
The test I provided is failing due to some other reason, let me fix the test.

pankajroark:
Updated the test in original comment.

ianoc:
Does this test pass or fail? You could open this as a simple unit test in a PR too thats failing.

pankajroark:
That test fails. Let me create a PR with the failing test.

johnynek:
honestly, naming and options are a weird, but we always felt like we could not fix them without breaking existing jobs.

We need a clear specification and more tests, I think.

pankajroark:
@johnynek curios what alternatives to options you considered. I don't think many of our users grok the way  application of settings via naming flows through the graph, so I agree options do seem weird to our users. Also agree that changing them without breaking existing jobs would be critical, so it's a tough problem. Still curios to know what your ideal solution would be.

johnynek:
well, we didn't think of removing options, but they were originally added when summingbird had no fan-out. We wanted a look of a builder:

``` scala
val myNode = fn(someStuff).name("myNode")
```

[...]

PR #672, opened 2016-06-29
Test to illustrate https://github.com/twitter/summingbird/issues/671

pankajroark:
To illustrate the issue with #671

### DIFF
--- a/summingbird-scalding-test/src/test/scala/com/twitter/summingbird/scalding/NamedOptionsSpec.scala
+++ b/summingbird-scalding-test/src/test/scala/com/twitter/summingbird/scalding/NamedOptionsSpec.scala
@@ -21,9 +21,10 @@ import cascading.pipe.Pipe
 import cascading.property.ConfigDef
 import cascading.property.ConfigDef.Setter
 import cascading.tuple.Fields
-import com.twitter.scalding.{ Test => TestMode, _ }
+import com.twitter.scalding.{Test => TestMode, _}
 import com.twitter.summingbird._
 import com.twitter.summingbird.batch.option.Reducers
+import com.twitter.summingbird.option.MonoidIsCommutative
 import org.scalatest.WordSpec
 
 /**
@@ -46,10 +47,17 @@ class NamedOptionsSpec extends WordSpec {
       override def apply(t: T) = t._1.toLong
     }
 
-  def pipeConfig(pipe: Pipe): Map[String, String] = {
+  def pipeConfig(pipe: Pipe): Map[String, List[String]] = {
     val configCollector = new Setter {
-      var config = Map.empty[String, String]
-      override def set(key: String, value: String): String = { config += key -> value; "" }
+      var config = Map.empty[String, List[String]]
+      override def set(key: String, value: String): String = {
+        if (config.contains(key)) {
+          config = config.updated(key, value :: config(key))
+        } else {
+          config += key -> List(value)
+        }
+        ""
+      }
       override def get(key: String): String = ???
       override def update(key: String, value: String): String = ???
     }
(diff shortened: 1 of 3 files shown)


johnynek:
we should see if this is also a bug with scalding.

We should have common code that does the naming based only on the Producer graph (and maybe on the items inside the graph that get the names, so we can actually translate the names).

I tried to do this in the past, but it was reverted.

summingbird-storm-test/src/test/scala/com/twitter/summingbird/storm/TopologyTests.scala
```diff
@@ -140,6 +124,23 @@ class TopologyTests extends WordSpec {
...
+      .sumByKey(TestStore.createStore[Int, Int]()._2).name(smNodeName)
+
+    val opts = Map(fmNodeName -> Options().set(SummerParallelism(10)),
+      smNodeName -> Options().set(SummerParallelism(20)))
+    val storm = Storm.local(opts)
+    val stormTopo = storm.plan(p).topology
+    val bolts = stormTopo.get_bolts
+
+    // Tail should use parallelism specified for the summer node
+    assert(bolts("Tail").get_common.get_parallelism_hint == 20)
```
jnievelt: Let's assert both settings?

pankajroark: Do you mean:
    assert(bolts("Tail").get_common.get_parallelism_hint != 10)
    assert(bolts("Tail").get_common.get_parallelism_hint == 20)

pankajroark: I'm trying to write the unit test for scalding platform. I'm trying to use the Reducers option for this. For the online platform tests we could inspect the created topology to check the number of summer nodes created, how do I do the corresponding thing for the scalding platform. i.e. how do I inspect the PipeFactory created after planning to check how many reducers it's going to use for  a node? I'm new to the scalding platform of summingbird, so may be missing obvious things, will appreciate any help. Thanks

pankajroark:
I added a few named option tests for scalding including for this case. This bug doesn't seem to affect scalding platform.

pankajroark:
Added a proposed fix, which is a single word change in StormPlatform's get function. The explanation is a bit complex though and follows:
The way node members are stored in SummerNode is initially Summer(IdentityKeyedProducer(FlatMappedProducer....))), IdentityKeyedProducer(FlatMappedProducer....))

but here https://github.com/twitter/summingbird/blob/develop/summingbird-online/src/main/scala/com/twitter/summingbird/planner/OnlinePlan.scala#L230 the order is reversed.

So node.members.last returns IdentityKeyedProducer(FlatMappedProducer....))

When the irrudicible analysis is done this IdentityKeyedProducer doesn't cover the Store used in the summer, and then during the reverse mapping it ends up getting mapped to both flatMapper and summer  named nodes.

Thus we end up looking at the options of both flatMapper and summer named nodes and find the flatMapper setting first.

By using node.members.head we use Summer(IdentityKeyedProducer(FlatMappedProducer....))) for look up and correctly find only the summer named node and thus only look up there.

NPraneeth:
While this fix solves this bug, i think it creates a new one. 

Lets suppose we have : 
    Source(**_).name("dummySrc").map ( fn ). name ("dummyOpt").sumByKey(**_*)
    Options.set("dummySrc"->SummerParallelism(10))
Which gets converted to :
    NamedProducer(OptionMappedProducer(NamedProducer(Source))) and the SourceNode will contain nodes : OptionMappedProducer, Source

When you take nodes.members.head -> you will get OptionMappedProuducer's options instead of the SourceProducer's options. 
So nodes.members.head might work in the summerNode but will be failing in the SourceNode. The same problem the other way around.

pankajroark:
@NPraneeth good catch.

summingbird-scalding-test/src/test/scala/com/twitter/summingbird/scalding/NamedOptionsSpec.scala
```diff
@@ -0,0 +1,217 @@
...
+        cfg.apply(ConfigDef.Mode.REPLACE, configCollector)
+      }
+      p.getPrevious.foreach(recurse(_))
+    }
+
+    recurse(pipe)
+    configCollector.config
+  }
+
+  def verify[T](
```
johnynek, line 77: what is T?

NPraneeth:
@pankajroark As the code for the 'get' method is in StormPlatform, we can use case statements to see if it's a SourceNode/SummerNode and try to collect the SourceProducer/SummerProducer and then grab the options corresponding to the producers.( This would be better than grabbing the options for the last node in the members ). And for FlatMapNode we can leave the node.members.last as default.
For Example : SummerNode -> Summer(IdentityKey(....
Instead of getting node.member.last ( IdentityKeyProd in this case ), we can identify it to be a SummerNode and grab the options for the SummerProducer instead. Similarly for the SourceNode too.

PR #674, opened 2016-07-18
Add some Execution support in scalding

johnynek:
This allows you to convert a `Producer[Scalding, T]` to a `Execution[TypedPipe[T]]` given a `DateRange` and optional options.

This does not convert summingbird to using Execution internally, but it maybe we can side-step that if we can safely use it from Execution (for the time being).


summingbird-scalding-test/src/test/scala/com/twitter/summingbird/scalding/ScaldingLaws.scala
```diff
@@ -131,11 +162,13 @@ class ScaldingLaws extends WordSpec {
       val summer = TestGraphs.singleStepJob[Scalding, (Long, Int), Int, Int](source, testStore)(t =>
         fn(t._2))
 
-      val scald = Scalding("scalaCheckJob")
-      val ws = new LoopState(intr)
+      val ex = Scalding.toExecutionExact(
```
johnynek: looks I screwed this up...

johnynek:
seems to pass tests for me now.

summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/ScaldingPlatform.scala
```diff
@@ -564,6 +566,44 @@ object Scalding {
...
+   */
+  def toExecutionExact[T](
+    dr: DateRange,
+    prod: Producer[Scalding, T],
+    options: Map[String, Options] = Map.empty): Execution[TypedPipe[(Timestamp, T)]] =
+
+    Execution.getMode.flatMap { mode =>
+      val fd = new FlowDef
+      toPipeExact(dr, prod, options)(fd, mode) match {
+        case Right(pipe) =>
```
rubanm: Had to double check if this was unused, so `Right(_)`?

rubanm:
👍

summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/ScaldingPlatform.scala
```diff
@@ -564,6 +566,44 @@ object Scalding {
...
+           * We plan twice. The first time we plan to see what new
+           * writes we do, the second time we only read, we do not
+           * do any new writes.
+           */
+          Execution.fromFn { (_, _) => fd }
+            .flatMap { _ =>
+              // Now plan again and use summingbird's built in support
+              // to minimize work by reading existing on-disk data:
+              val reads = new FlowDef
+              toPipeExact(dr, prod, options)(reads, mode) match {
```
ulyssepence, line 590: Forgive me, I'm very unfamiliar with the SB codebase. Don't we have the expectation that `Config` options provided to an `Execution` could affect the behavior of how the Scalding Job runs? Or is that supposed to be represented in `options`?

johnynek, line 590: Config is passed to `Execution`. `Options` are summingbird specific and are passed to `toPipeExact`.

We don't have the expectation that scalding options can control summingbird planning.

ulyssepence, line 590: OK sounds good.

PR #684, opened 2016-08-26
Create COMMITTERS.md

isnotinvain:


ulyssepence:
+1

sritchie:
@isnotinvain do you mind adding me to this list as well? I haven't committed in a while, of course, but that'll change soon!

isnotinvain:
Hey! So the way we're thinking about the list of COMMITTERS is along the lines of apache project committers (so this is not a list of authors, nor a list of contributors). We've written up something for our process here (yay process): https://github.com/twitter/analytics-infra-governance

For this first round of seeding the initial committers (eg, folks who can +1 and merge a PR), we went with the twitter CDL team + Ian and Oscar because they've been acting as project committers within the last 1 year. We knew for sure there are way more people who've been past committers / project maintainers and didn't really want to tackle that in this first pass, so that's why you weren't included.

Would you like to be a committer (do code reviews, fix tech debt, do project maintenance stuff) going forward? We haven't really tried this out yet, so maybe we could use you as a test case and go through the voting process outlined in https://github.com/twitter/analytics-infra-governance ?

We can also add you to the emeritus list for sure if you don't want to sign up for the project maintenance side of things.

What do you think?

PR #685, opened 2016-09-02
Extracting Future queuing logic from AsyncBase

jnievelt:
Adding `FutureQueue`, a structure that essentially takes in `(S, Future[T])` and gives back `(S, Try[T])`.

I didn't mean to totally gut `AsyncBase`, but it's mostly become a shim integrating its own interface (`apply`, `tick`), the `OperationContainer` interface (`execute`, `executeTick`), and its `FutureQueue`.

Making this private for now.

Adding a few more tests while I'm at it.  Cleaning up Node.scala style since it does that automatically.

The next step would be to integrate this with `AggregatingOutputCollector`.

### DIFF
--- a/summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala
+++ b/summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala
@@ -19,7 +19,7 @@ package com.twitter.summingbird.online.executor
 import java.util.concurrent.TimeoutException
 import java.util.concurrent.atomic.AtomicInteger
 
-import com.twitter.summingbird.online.Queue
+import com.twitter.summingbird.online.FutureQueue
 import com.twitter.summingbird.online.option.{ MaxEmitPerExecute, MaxFutureWaitTime, MaxWaitingFutures }
 import com.twitter.util._
 
@@ -35,21 +35,19 @@ abstract class AsyncBase[I, O, S, D, RC](maxWaitingFutures: MaxWaitingFutures, m
   def apply(state: S, in: I): Future[TraversableOnce[(Seq[S], Future[TraversableOnce[O]])]]
   def tick: Future[TraversableOnce[(Seq[S], Future[TraversableOnce[O]])]] = Future.value(Nil)
 
-  private[executor] lazy val futureQueue = new FutureQueue[Seq[S], TraversableOnce[O]](maxWaitingFutures, maxWaitingTime, maxEmitPerExec)
+  private[executor] lazy val futureQueue = new FutureQueue[Seq[S], TraversableOnce[O]](maxWaitingFutures, maxWaitingTime)
 
-  override def executeTick =
-    finishExecute(Nil, tick)
+  override def executeTick: TraversableOnce[(Seq[S], Try[TraversableOnce[O]])] =
+    finishExecute(None, tick)
 
-  override def execute(state: S, data: I) =
-    finishExecute(List(state), apply(state, data))
+  override def execute(state: S, data: I): TraversableOnce[(Seq[S], Try[TraversableOnce[O]])] =
+    finishExecute(Some(state), apply(state, data))
 
-  private def finishExecute(states: Seq[S], fIn: Future[TraversableOnce[(Seq[S], Future[TraversableOnce[O]])]]) = {
-    futureQueue.addAllFuture(states, fIn)
-
-    // always empty the responses
-    futureQueue.dequeue
+  private def finishExecute(failStateOpt: Option[S], fIn: Future[TraversableOnce[(Seq[S], Future[TraversableOnce[O]])]]) = {
+    fIn.respond {
+      case Return(iter) => futureQueue.addAll(iter)
+      case Throw(ex) => futureQueue.add(failStateOpt.toSeq, Future.exception(ex))
+    }
+    futureQueue.dequeue(maxEmitPerExec.get)
   }
-
-  private def handleSuccess(fut: Future[TraversableOnce[(Seq[S], Future[TraversableOnce[O]])]]) =
-    fut.onSuccess { iter => futureQueue.addAll(iter) }
 }
--- a/summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala
+++ b/summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala
@@ -22,49 +22,10 @@ import java.util.concurrent.atomic.AtomicInteger
 import com.twitter.summingbird.online.Queue
 import com.twitter.summingbird.online.option.{ MaxEmitPerExecute, MaxFutureWaitTime, MaxWaitingFutures }
 import com.twitter.util._
-import org.slf4j.{ Logger, LoggerFactory }
 
 import scala.util.{ Failure, Success, Try }
 
-object AsyncBase {
-  /**
-   * Ratio of total number of outstanding futures to the portion that is finished
-   * ,at which finished futures are cleared.
-   * Clearing finished futures costs proportional to total number of outstanding
-   * futures, so we want to make sure we only clear when sufficient portion is
-   * finished.
-   */
-  val OutstandingFuturesDequeueRatio = 2
-
-  /**
-   * Wait for n futures to finish. Doesn't block, the returned future is satisfied
-   * once n futures have finished either successfully or unsuccessfully.
-   * If n is greater than number of futures in queue then we wait on all of them.
-   */
-  def waitN[A](fs: Iterable[Future[A]], n: Int): Future[Unit] = {
-    val waitOnCount = Math.min(fs.size, n)
-    if (waitOnCount <= 0) {
-      Future.Unit
-    } else {
-      val count = new AtomicInteger(waitOnCount)
-      val p = Promise[Unit]()
-      fs.foreach { f =>
-        f.ensure {
-          // Note that since we are only decrementing we can cross 0 only
-          // once (unless we decrement more than 2^32 times).
-          if (count.decrementAndGet() == 0) {
-            p.setValue(())
-          }
-        }
-      }
-      p
-    }
-  }
-}
-
 abstract class AsyncBase[I, O, S, D, RC](maxWaitingFutures: MaxWaitingFutures, maxWaitingTime: MaxFutureWaitTime, maxEmitPerExec: MaxEmitPerExecute) extends Serializable with OperationContainer[I, O, S, D, RC] {
-
-  @transient protected lazy val logger: Logger = LoggerFactory.getLogger(getClass)
 
   /**
    * If you can use Future.value below, do so. The double Future is here to deal with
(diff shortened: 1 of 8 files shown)


summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/FutureQueue.scala
```diff
@@ -0,0 +1,150 @@
...
+            p.setValue(())
+          }
+        }
+      }
+      p
+    }
+  }
+}
+
+private[summingbird] class FutureQueue[S, T](
```
johnynek: can we comment the type parameters?

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/FutureQueue.scala
```diff
@@ -0,0 +1,150 @@
...
+    maxWaitingFutures: MaxWaitingFutures,
+    maxWaitingTime: MaxFutureWaitTime,
+    maxEmitPerExec: MaxEmitPerExecute) {
+  @transient protected lazy val logger: Logger = LoggerFactory.getLogger(getClass)
+
+  private[executor] lazy val outstandingFutures = Queue.linkedNonBlocking[Future[Unit]]
```
johnynek: can this just be a regular `private`?

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/FutureQueue.scala
```diff
@@ -0,0 +1,150 @@
...
+  private lazy val numPendingOutstandingFutures = new AtomicInteger(0)
+  private lazy val responses = Queue.linkedNonBlocking[(S, Try[T])]
+
+  def addAll(iter: TraversableOnce[(S, Future[T])]): Unit = {
+    val addedSize = iter.foldLeft(0) {
+      case (size, (state, fut)) =>
+        val responded =
+          fut
```
johnynek: what about using `fut.transform { t => responses.put((state, t)); Future.const(t) }` instead of `onSuccess` and `onFailure`?

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/FutureQueue.scala
```diff
@@ -0,0 +1,150 @@
...
+    maxWaitingTime: MaxFutureWaitTime,
+    maxEmitPerExec: MaxEmitPerExecute) {
+  @transient protected lazy val logger: Logger = LoggerFactory.getLogger(getClass)
+
+  private[executor] lazy val outstandingFutures = Queue.linkedNonBlocking[Future[Unit]]
+  private lazy val numPendingOutstandingFutures = new AtomicInteger(0)
+  private lazy val responses = Queue.linkedNonBlocking[(S, Try[T])]
+
+  def addAll(iter: TraversableOnce[(S, Future[T])]): Unit = {
+    val addedSize = iter.foldLeft(0) {
```
johnynek: what about:

``` scala
val addedSize = iter.map { case (state, fut) =>
  val responded = ...
  if (addOutstandingFuture(responded)) 1 else 0
}.sum 
```

johnynek: or, if we care about perf:

``` scala
var addedSize = 0
iter.foreach { case (state, fut) =>
  ...
  if ( ) addedSize += 1
}
```

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/FutureQueue.scala
```diff
@@ -0,0 +1,150 @@
...
+       * May indicate maxWaitingFutures is too low.
+       */
+      logger.debug(
+        "Exceeded maxWaitingFutures({}), put {} futures", maxWaitingFutures.get, addedSize
+      )
+    }
+  }
+
+  def addAllFuture(state: S, iterFut: Future[TraversableOnce[(S, Future[T])]]): Unit =
+    addOutstandingFuture(
```
johnynek: can we use `transform` here as well. Not so crazy about the `.onSuccess` `onFailure` while we are making changes.

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/FutureQueue.scala
```diff
@@ -0,0 +1,150 @@
...
+    maxWaitingFutures: MaxWaitingFutures,
+    maxWaitingTime: MaxFutureWaitTime,
+    maxEmitPerExec: MaxEmitPerExecute) {
+  @transient protected lazy val logger: Logger = LoggerFactory.getLogger(getClass)
+
+  private[executor] lazy val outstandingFutures = Queue.linkedNonBlocking[Future[Unit]]
+  private lazy val numPendingOutstandingFutures = new AtomicInteger(0)
+  private lazy val responses = Queue.linkedNonBlocking[(S, Try[T])]
+
+  def addAll(iter: TraversableOnce[(S, Future[T])]): Unit = {
```
johnynek: I'm not sure we should have this method. Why not:

``` scala
def add(s: S, v: Future[T]): Boolean =
```

then when we need to call `addAll` instead do: `val addedSize = iter.count { case (s, fut) => add(s, fut) }` and optionally log there?

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/FutureQueue.scala
pankajroark: Can you also add a doc comment describing what the semantics are? Especially around blocking and how State relates to Future.

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/FutureQueue.scala
jnievelt: Tests are currently inspecting this.  It might make sense to modify/eliminate that test (I've broken it for now anyway, I think).  Otherwise it can be private.

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/FutureQueue.scala
jnievelt: I can use `Future#respond`, but it won't be quite as clean as that because we're also converting between Twitter & Scala `Try`.

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/FutureQueue.scala
```diff
@@ -0,0 +1,150 @@
...
+       * This can happen on large key expansion.
+       * May indicate maxWaitingFutures is too low.
+       */
+      logger.debug(
+        "Exceeded maxWaitingFutures({}), put {} futures", maxWaitingFutures.get, addedSize
+      )
+    }
+  }
+
+  def addAllFuture(state: S, iterFut: Future[TraversableOnce[(S, Future[T])]]): Unit =
```
pankajroark: Can you document the state contract here? Looks like "state" is only used for putting into failure response. Looking at FutureQueue as a standalone abstraction what is the reasoning behind such api design. Perhaps we should rename the parameter to reflect how it's used.

summingbird-online/src/test/scala/com/twitter/summingbird/online/executor/FutureQueueLaws.scala
```diff
@@ -0,0 +1,148 @@
...
+}
+
+class FutureQueueLaws extends Properties("FutureQueue") with Eventually {
+  def genTry[T](implicit arb: Arbitrary[T]): Gen[Try[T]] = Gen.oneOf(arb.arbitrary.map(Return(_)), Arbitrary.arbitrary[java.lang.Exception].map(Throw(_)))
+  implicit def arbTry[T: Arbitrary] = Arbitrary(genTry[T])
+
+  implicit val arbNonNegativeShort: Arbitrary[NonNegativeShort] = Arbitrary(
+    Arbitrary.arbitrary[Short].filter { _ >= 0 }.map { NonNegativeShort }
+  )
+  def twitterToScala[T](t: Try[T]) = t match {
```
johnynek: I've seen this written like 100 times. I finally got it into util, but we might not want to upgrade the dependency just to get it (we are having some GC regressions in some of the latest finagles).

jnievelt: Looks like we already pick up bijection-util so I can just use what we have in `UtilBijections`.

jnievelt:
These comments should be addressed.  On review, there is sort of a dependency on `Queue`, so maybe the dream of moving this out of summingbird is a bit farfetched?

I'm thinking about moving it up into the `online` package next to `Queue`.  Mixed feelings about leaving is private vs. not.

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/FutureQueue.scala
```diff
@@ -141,6 +188,11 @@ private[summingbird] class FutureQueue[S, T](
     }
   }
 
+  /**
+   * Retrieve any completed results, along with their associated states.
+   *
+   * Returns up to maxEmitPerExec items.
+   */
   def dequeue: TraversableOnce[(S, Try[T])] = {
```
johnynek: just a question, can't this be `Iterable[(S, Try[T])]`? `.take` returns something iterable, right? This seems to limit the caller to only go through it once, but in fact, we materialize it into memory already.

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/FutureQueue.scala
```diff
@@ -0,0 +1,202 @@
...
+    if (!fut.isDefined) {
+      numPendingOutstandingFutures.incrementAndGet
+      val ensured = fut.ensure(numPendingOutstandingFutures.decrementAndGet)
+      outstandingFutures.put(ensured)
+      true
+    } else {
+      false
+    }
+
+  private def forceExtraFutures() {
```
johnynek: can we make this `forceExtraFutures(): Unit = {` I think that is the preferred scala style now.

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/FutureQueue.scala
```diff
@@ -0,0 +1,202 @@
...
+      }
+    }
+  }
+
+  /**
+   * Retrieve any completed results, along with their associated states.
+   *
+   * Returns up to maxEmitPerExec items.
+   */
+  def dequeue: TraversableOnce[(S, Try[T])] = {
```
johnynek: can we return `Iterable[(S, Try[T])]` here?

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/FutureQueue.scala
```diff
@@ -0,0 +1,202 @@
...
+    // don't let too many futures build up
+    forceExtraFutures()
+    // Take all results that have been placed for writing to the network
+    responses.take(maxEmitPerExec.get)
```
johnynek: can we not take `maxEmitPerExec` as a class param and instead take it as a method arg here:

``` scala
def dequeue(max: Int): Iterable[(S, Try[T])] = ...
```

This simplifies the state of the class and allows the caller to have better control (no need for a uniform max).

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/FutureQueue.scala
```diff
@@ -0,0 +1,202 @@
...
+ *
+ * To support additional asynchronous behavior, a state value along with a
+ * Future[TraversableOnce[(S, Future[T])]] can be inserted.  On failure, the
+ * outer state value is returned with the Failure.  On success, the inner
+ * items are queued up to be inserted when they complete.
+ */
+private[summingbird] class FutureQueue[S, T](
+    maxWaitingFutures: MaxWaitingFutures,
+    maxWaitingTime: MaxFutureWaitTime,
+    maxEmitPerExec: MaxEmitPerExecute) {
```
johnynek: can we remove this param and pass it to `dequeue`?

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/FutureQueue.scala
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  /**
+   * Queue a collection Future.  On failure, the state can be retrieved
+   * with the failure.  On success, the results are queued via addAll.
+   *
+   * The Future given here counts against the maxWaitingFutures, so calls
+   * to this method may cause a wait on dequeue.
+   */
+  def addAllFuture(state: S, iterFut: Future[TraversableOnce[(S, Future[T])]]): Unit =
```
johnynek: this seems like a really weird method. I think in the original, with `S = Seq[S1]` then what you actually have is `state` is the union of all states, where each part is separate in `TraversableOnce`.

So, really, I think you have something like, there is a `Monoid[S]` and the sum of all the `TraversableOnce[(S, _)]` is the state.

So, I don't see how this can be general without it.

In the general case, I think we might want something like `Seq[S], Future[Seq[Future[T]]]` so we have a sequence in the same order of states and `T`s. Is that right?

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala
```diff
@@ -74,92 +35,21 @@ abstract class AsyncBase[I, O, S, D, RC](maxWaitingFutures: MaxWaitingFutures, m
...
-    finishExecute(tick.onFailure { thr => responses.put(((Seq(), Failure(thr)))) })
+    finishExecute(Nil, tick)
 
   override def execute(state: S, data: I) =
-    finishExecute(apply(state, data).onFailure { thr => responses.put(((List(state), Failure(thr)))) })
+    finishExecute(List(state), apply(state, data))
 
-  private def finishExecute(fIn: Future[TraversableOnce[(Seq[S], Future[TraversableOnce[O]])]]) = {
-    addOutstandingFuture(handleSuccess(fIn).unit)
+  private def finishExecute(states: Seq[S], fIn: Future[TraversableOnce[(Seq[S], Future[TraversableOnce[O]])]]) = {
```
johnynek: actually, it looks like `Seq[S]` is really an Option, is that right? Can we simplify this?

johnynek: actually, only the first `Seq[S]` is an option, so that breaks the uniformity of the typing in FutureQueue.

I wonder if `addAllFuture` is so specialized, it should stay in this class. I don't see how any user can use it generally without a big coincidence in the `FutureQueue` implementation. To me, I think so, then I think `FutureQueue` becomes very clear and simple.

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/FutureQueue.scala
```diff
@@ -0,0 +1,202 @@
...
+  // When futures complete, they deposit their results here.
+  private lazy val responses = Queue.linkedNonBlocking[(S, Try[T])]
+
+  private val tryBijection = UtilBijections.twitter2ScalaTry[T]
+
+  /**
+   * Add a Future with state onto the queue.  This can eventually be
+   * dequeued as (S, Try[T]) when the Future completes.
+   *
+   * Returns true if the Future is not yet complete.  When false, the
```
pankajroark: The meaning of false return value seems confusing. Why would caller of add care whether that future counts towards maxWaitingFutures limit or not; because of asynchronous execution user doesn't have control over how futures finish and thus can't do anything about it when encountering a false value. 

For addOutstandingFutures the return value of false is ok, since that is internal to this class and false more directly indicates that we could not add to the list of outstanding futures. But I don't see much value in returning that as the return value of add.

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/FutureQueue.scala
jnievelt: Yeah I guess `Queue` is set to return `Seq`; seems reasonable to do the same here.

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala
jnievelt: Sure, I like that approach too.

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/FutureQueue.scala
jnievelt: Fair point.  I was anyway a little worried that folks would confuse this with "item was queued or not" semantics.  Making this a `Unit` method simplifies that; I'll make this a private internal method and making a public wrapper for it.

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala
jnievelt: On review, the main thing we lose is the presence of the outer `Future` on the queue (as something counting against the max that we can wait on, etc.).  Maybe this is a reasonable tradeoff though.

jnievelt:
Any remaining concerns on this one?  I've verified that the current diff works fine with my planned changes to `AggregatorOutputCollector`.

johnynek:
+1 this is really great work, @jnievelt 

I think the `FutureQueue` is really a general rate-limiter on `S => Future[T]` now.

johnynek (approved):

PR #686, opened 2016-09-06
Add Sam Ritchie as a COMMITTER

isnotinvain:
I think it was an oversight to not include Sam as one of the original COMMITTERs in summingbird, so this PR adds him.

I don't think a vote is necessary, as we did not vote on the original set of COMMITTERS, but after this, I think we can consider the original 'grandfathered' committers to be closed and all other committers should go through a vote (which is not a difficult process, so if we missed somebody obvious, we can vote, it'll pass, and that'll be that).


piyushnarang:
👍

isnotinvain:
Are com.twitter.summingbird.storm.StormLaws known to be flakey? They seemed to have failed.

isnotinvain:
Also given how long these tests run, I wonder if we can disable whatever causes PRs to run tests twice (seems to run them because they are PRs and also because they are branches?)

johnynek:
@isnotinvain yes, stormlaws are flakey. See #605

johnynek:
👍

sritchie:
thanks all!

PR #687, opened 2016-09-08
Pass the semigroup to create Mergeable online

### LINKED ISSUE #677: Storm Store type should take `Semigroup[V]`
Right now, the producer graph has a `Semigroup[V]` in `sumByKey`, but actually, the storm platform ignores it.

Scalding does not.

https://github.com/twitter/summingbird/blob/develop/summingbird-online/src/main/scala/com/twitter/summingbird/online/MergeableStoreFactory.scala

A fix for this is to have `MergeableStoreFactory` really be something like `Semigroup[V] => Mergeable[K, V]` rather than `() => Mergeable[K, V]`.

We can be somewhat backwards compatible by giving the old methods (but marking them as deprecated) and just ignoring the `Semigroup` passed in.

Without something like this #562 is pretty dangerous because online is would be really easy to have a mismatch between the store semigroup and the aggregator semigroup (this is actually already pretty dangerous and not clear how we didn't address this before now).

johnynek:
closes #677

The issue is that it is easy to have the semigroup get out of sync between the `implicit` semigroup in the `Producer` graph, and the captured value for storm. This is not totally fixed, since when using a single store as a service and a store, we still need to capture the semigroup, but `leftJoin` does not have access to that. We could fix that walking the graph and finding the semigroup at the location of the `sumByKey`. We can do that in a later PR.

### DIFF
--- a/summingbird-online/src/main/scala/com/twitter/summingbird/online/CombinedServiceStoreFactory.scala
+++ b/summingbird-online/src/main/scala/com/twitter/summingbird/online/CombinedServiceStoreFactory.scala
@@ -28,22 +28,19 @@ object CombinedServiceStoreFactory {
   def apply[K, V](onlineStore: => MergeableStore[(K, BatchID), V], batchesToKeep: Int)(implicit b: Batcher) = {
 
     lazy val clientStore = ClientStore[K, V](onlineStore, batchesToKeep)(b, onlineStore.semigroup)
-
-    new CombinedServiceStoreFactory[K, V] {
-      def mergeableStore = () => onlineStore
-      def mergeableBatcher = b
-      def serviceStore = () => clientStore
-    }
+    from[K, V](clientStore, onlineStore, b)
   }
 
   def apply[K, V](offlineStore: => ReadableStore[K, (BatchID, V)], onlineStore: => MergeableStore[(K, BatchID), V], batchesToKeep: Int)(implicit b: Batcher) = {
 
     lazy val clientStore = ClientStore[K, V](offlineStore, onlineStore, batchesToKeep)(b, onlineStore.semigroup)
-
-    new CombinedServiceStoreFactory[K, V] {
-      def mergeableStore = () => onlineStore
-      def mergeableBatcher = b
-      def serviceStore = () => clientStore
-    }
+    from[K, V](clientStore, onlineStore, b)
   }
+
+  private[this] def from[K, V](cs: => ReadableStore[K, V], online: => MergeableStore[(K, BatchID), V], b: Batcher): CombinedServiceStoreFactory[K, V] =
+    new CombinedServiceStoreFactory[K, V] {
+      def mergeableStore = _ => online
+      def mergeableBatcher = b
+      def serviceStore = () => cs
+    }
 }
--- a/summingbird-online/src/main/scala/com/twitter/summingbird/online/MergeableStoreFactory.scala
+++ b/summingbird-online/src/main/scala/com/twitter/summingbird/online/MergeableStoreFactory.scala
@@ -39,10 +39,8 @@ object MergeableStoreFactory {
       def mergeableBatcher = batcher
     }
 
-  def fromOnlineOnly[K, V](store: => MergeableStore[K, V]): MergeableStoreFactory[(K, BatchID), V] = {
-    implicit val batcher = Batcher.unit
-    from(store.convert { k: (K, BatchID) => k._1 })
-  }
+  def fromOnlineOnly[K, V](store: => MergeableStore[K, V]): MergeableStoreFactory[(K, BatchID), V] =
+    fromOnlineOnlyWithSemigroup(_ => store)
 
   def fromOnlineOnlyWithSemigroup[K, V](fn: Semigroup[V] => MergeableStore[K, V]): MergeableStoreFactory[(K, BatchID), V] =
     fromWithSemigroup(fn.andThen(_.convert { k: (K, BatchID) => k._1 }))(Batcher.unit)
--- a/summingbird-online/src/main/scala/com/twitter/summingbird/online/MergeableStoreFactory.scala
+++ b/summingbird-online/src/main/scala/com/twitter/summingbird/online/MergeableStoreFactory.scala
@@ -16,6 +16,7 @@
 
 package com.twitter.summingbird.online
 
+import com.twitter.algebird.Semigroup
 import com.twitter.storehaus.algebra.{ MergeableStore, Mergeable, StoreAlgebra }
 import com.twitter.summingbird.batch.{ Batcher, BatchID }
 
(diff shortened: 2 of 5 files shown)


summingbird-online/src/main/scala/com/twitter/summingbird/online/CombinedServiceStoreFactory.scala
```diff
@@ -28,22 +28,19 @@ object CombinedServiceStoreFactory {
...
   def apply[K, V](offlineStore: => ReadableStore[K, (BatchID, V)], onlineStore: => MergeableStore[(K, BatchID), V], batchesToKeep: Int)(implicit b: Batcher) = {
 
     lazy val clientStore = ClientStore[K, V](offlineStore, onlineStore, batchesToKeep)(b, onlineStore.semigroup)
+    from[K, V](clientStore, onlineStore, b)
+  }
 
+  private[this] def from[K, V](cs: => ReadableStore[K, V], online: => MergeableStore[(K, BatchID), V], b: Batcher): CombinedServiceStoreFactory[K, V] =
     new CombinedServiceStoreFactory[K, V] {
-      def mergeableStore = () => onlineStore
+      def mergeableStore = _ => online
```
jnievelt, line 42: Inconsistency between `_` and `()`

johnynek: what do you mean? They mean different things. `_: Semigroup[V]` is being ignored here.

jnievelt: Ah, of course

summingbird-online/src/main/scala/com/twitter/summingbird/online/MergeableStoreFactory.scala
```diff
@@ -26,23 +27,28 @@ import com.twitter.summingbird.batch.{ Batcher, BatchID }
...
       def mergeableBatcher = batcher
     }
-  }
-
-  def from[K, V](store: => Mergeable[(K, BatchID), V])(implicit batcher: Batcher): MergeableStoreFactory[(K, BatchID), V] =
-    apply({ () => store }, batcher)
 
   def fromOnlineOnly[K, V](store: => MergeableStore[K, V]): MergeableStoreFactory[(K, BatchID), V] = {
     implicit val batcher = Batcher.unit
     from(store.convert { k: (K, BatchID) => k._1 })
```
jnievelt: `fromOnlineOnlyWithSemigroup(_ => store)` ?

johnynek: good call.

jnievelt:
Looks good.  I suppose we'll need a version bump for the `MergeableStoreFactory` signature change?  At least it should be a trivial modification to make sources compatible.

johnynek:
@jnievelt yes, this will be a minor version break.

summingbird-online/src/main/scala/com/twitter/summingbird/online/MergeableStoreFactory.scala
```diff
@@ -26,23 +27,26 @@ import com.twitter.summingbird.batch.{ Batcher, BatchID }
...
-    implicit val batcher = Batcher.unit
-    from(store.convert { k: (K, BatchID) => k._1 })
-  }
+  def fromOnlineOnlyWithSemigroup[K, V](fn: Semigroup[V] => MergeableStore[K, V]): MergeableStoreFactory[(K, BatchID), V] =
+    fromWithSemigroup(fn.andThen(_.convert { k: (K, BatchID) => k._1 }))(Batcher.unit)
 }
 
 trait MergeableStoreFactory[-K, V] extends java.io.Serializable {
-  def mergeableStore: () => Mergeable[K, V]
+  def mergeableStore: Semigroup[V] => Mergeable[K, V]
```
pankajroark, line 50: At this point it seems like this method should be named like a function and not a property. Earlier it was in the spirit of  lazily created mergeable store so the name was ok. Should we call it createMergeableStore?

pankajroark:
Looks good overall, one minor comment.

pankajroark:
@johnynek Could you cut a release before merging this? This will for sure require changes in our internal repository since there are custom implementations of MergeableStoreFactory that will need to change.

summingbird-online/src/main/scala/com/twitter/summingbird/online/MergeableStoreFactory.scala
johnynek: Well, I'm a little nervous about that because it lead to a lot of serialization problems before (which is why we did this). Calling a method is still lazy, so `def mergeableStore = store` would have been lazy, but still we got burned enough to add this extra layer).

johnynek:
@pankajroark I want to push back on cutting a release. The reason is this:
1. I don't want to see us raise the bar to "any binary change should have a published release before merging"
2. this change is:
   a. unlikely to hit many users I assume (who should have used the constructor functions in most cases)
   b. totally mechanical to port (just ignore the semigroup for the old behavior, or actually use it to get a safer operation).
3. The work would be throw-away, since we would need to publish again (since we publish RCs from develop), and any work on a merge would need to be done against the new jar.
4. We can always roll-back if there is some reason why this change is too onerous (but fixing code to be safer, seems to me like generally a win, not a cost). In fact, lots of the reason for custom MergeableStoreFactory is because you need to know exactly what semigroup is. With this change, that is no longer needed, so I imagine you can have maybe just one or two implementations throughout Twitter.

To turn it around: should Twitter employees publish new jars on each incompatible change for other users to see what the change will be like before a merge? I think that is too much work.

jnievelt:
Actually I think there's just one place where we call `MergeableStoreFactory#mergeableStore` and it's in a test.  The constructions (using `from` and `apply`) should be source compatible at least.  I don't think it's a problem to merge before releasing.

summingbird-online/src/main/scala/com/twitter/summingbird/online/MergeableStoreFactory.scala
pankajroark: I didn't mean to suggest we change the structure, just the name. MergeableStoreFactory.mergeableStore reads like a property which would provide a mergeable store but it is a function that takes an input to provide a mergeable store. 

e.g. when I read factory.mergeableStore I would think that this is a mergeable store object, finding out that it is a function that needs a semigroup as input to produce a mergeable store would be a surprise.

pankajroark:
Yeah, this should be ok. Overriding Mergeable is common, MergeableStoreFactory seems to be created mostly with the help of factory function. By bad in confusing the two.

On cutting a release on breaking changes, it was just an effort on my part to make my life easier. I'm planning to do a summingbird release soon and was hoping to do least amount of code changes in our repo, the last release turned out to be very complicated due to the MemoryPlatform changes. But it's not a big deal. In general I agree I wouldn't expect a release of summingbird on every breaking change.

johnynek:
I have a +1 on this now, @pankajroark @jnievelt ?

jnievelt (approved):

PR #693, opened 2016-10-07
Aggregate message IDs with values in AggregatorOutputCollector

### LINKED ISSUE #692: AggregatorOutputCollector emits too many message IDs
The current logic is something like:
- add things to an `AsyncSummer`
- anything that comes out of the `addAll` call is converted to the (keyIndex, (key, value)) format
- for every keyIndex, send out all the message IDs we have pending for that index

Of course, if we have two colliding keys being summed, and an `AsyncSummer` implementation that doesn't do all-or-nothing flushes on `addAll`/etc., an emit of one key would include the message IDs for the aggregations on the other key as well.

I think I missed this in review, writing it off as okay because we are doing synchronous handling on the `addAll` result.  But actually we still need to track the message IDs along with the aggregated values, a la `FinalFlatMap` and `Summer`.

I have the fix ready, originally thinking the change was only needed to do Future queuing, but I can split it out since they're really independent.

jnievelt:
Fixes #692

Getting rid of `streamMessageIdTracker` because that's the wrong scope to track message IDs.  Pushing the message IDs instead into the `AsyncSummer`.

Reworking `TestAggregateOutpoutCollector` a bit to accept expected message IDs and stream names.  Existing tests just use `None` for these.

Adding new `AggregatorOutputCollectorTest`.  This is pretty simple, using an `AsyncSummer` that always returns the same thing from `addAll`, but is sufficient to illustrate the issue captured in #692.

### DIFF
--- a/summingbird-storm/src/main/scala/com/twitter/summingbird/storm/collector/AggregatorOutputCollector.scala
+++ b/summingbird-storm/src/main/scala/com/twitter/summingbird/storm/collector/AggregatorOutputCollector.scala
@@ -37,7 +37,7 @@ class AggregatorOutputCollector[K, V: Semigroup](
       if (iter.isEmpty) None else Some(iter.flatten)
   }
 
-  private def getSummer = summerBuilder.getSummer[K, (OutputMessageId, V)](implicitly[Semigroup[(OutputMessageId, V)]])
+  private def getSummer = summerBuilder.getSummer[K, (OutputMessageId, V)]
 
   /**
    * This method is invoked from the nextTuple() of the spout.
--- a/summingbird-storm/src/main/scala/com/twitter/summingbird/storm/collector/AggregatorOutputCollector.scala
+++ b/summingbird-storm/src/main/scala/com/twitter/summingbird/storm/collector/AggregatorOutputCollector.scala
@@ -56,10 +56,10 @@ class AggregatorOutputCollector[K, V: Semigroup](
   private def convertToSummerInputFormat(flushedCache: Map[K, (OutputMessageId, V)]): TraversableOnce[OutputTuple] =
     flushedCache.groupBy {
       case (k, _) => summerShards.summerIdFor(k)
-    }.map {
-      case (index: AggKey, m: Map[K, (OutputMessageId, V)]) =>
-        val messageIds = m.values.iterator.flatMap { case (ids, _) => ids }
-        val results = m.mapValues { case (_, v) => v }
+    }.iterator.map {
+      case (index, data) =>
+        val messageIds = data.values.iterator.flatMap { case (ids, _) => ids }
+        val results = data.mapValues { case (_, v) => v }
         (index, results, messageIds)
     }
 
--- a/summingbird-storm/src/main/scala/com/twitter/summingbird/storm/collector/AggregatorOutputCollector.scala
+++ b/summingbird-storm/src/main/scala/com/twitter/summingbird/storm/collector/AggregatorOutputCollector.scala
@@ -25,11 +25,19 @@ class AggregatorOutputCollector[K, V: Semigroup](
 
   private type AggKey = Int
   private type AggValue = Map[K, V]
-  private type OutputMessageId = TraversableOnce[Object]
+  private type OutputMessageId = Iterator[Object]
   private type OutputTuple = (AggKey, AggValue, OutputMessageId)
 
   // An individual summer is created for each stream of data. This map keeps track of the stream and its corresponding summer.
-  private val cacheByStreamId = MMap.empty[String, AsyncSummer[(K, (Seq[Object], V)), Map[K, (Seq[Object], V)]]]
+  private val cacheByStreamId = MMap.empty[String, AsyncSummer[(K, (OutputMessageId, V)), Map[K, (OutputMessageId, V)]]]
+
+  implicit val messageIdSg = new Semigroup[OutputMessageId] {
+    override def plus(a: OutputMessageId, b: OutputMessageId) = a ++ b
+    override def sumOption(iter: TraversableOnce[OutputMessageId]) =
+      if (iter.isEmpty) None else Some(iter.flatten)
+  }
+
+  private def getSummer = summerBuilder.getSummer[K, (OutputMessageId, V)](implicitly[Semigroup[(OutputMessageId, V)]])
 
   /**
    * This method is invoked from the nextTuple() of the spout.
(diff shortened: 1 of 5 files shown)


summingbird-storm/src/main/scala/com/twitter/summingbird/storm/collector/AggregatorOutputCollector.scala
```diff
@@ -46,27 +45,33 @@ class AggregatorOutputCollector[K, V: Semigroup](
     flushExecTimeCounter.incrBy(Time.now.inMillis - startTime.inMillis)
   }
 
-  private def convertToSummerInputFormat(flushedCache: CMap[_, _]): CMap[Int, CMap[_, _]] =
-    flushedCache.groupBy { case (k, _) => summerShards.summerIdFor(k) }
+  private def convertToSummerInputFormat(flushedCache: Map[K, (Seq[Object], V)]): TraversableOnce[OutputTuple] =
+    flushedCache.groupBy {
+      case (k, _) => summerShards.summerIdFor(k)
+    }.map {
+      case (index: AggKey, m: Map[K, (OutputMessageId, V)]) =>
```
johnynek: `case f: Foo` is a pattern match on type. Can we remove the types to be clear we are not doing that? If you want to annotate, you can do in the next line, or do:

``` scala
.map { tup: (AggKey, Map[...]) =>
  val (index, m) = tup
```

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/collector/AggregatorOutputCollector.scala
```diff
@@ -75,43 +80,24 @@ class AggregatorOutputCollector[K, V: Semigroup](
...
-   * The lookup of stream is followed by the lookup on summerShard happens.
-   * The messageId is added to the corresponding group.
-   * All the messageIds are sent along with the crushed tuple.
-   */
-  private def trackMessageId(tuple: (K, V), o: AnyRef, s: String): Unit = {
-    val messageIdTracker = streamMessageIdTracker.getOrElseUpdate(s, MMap[Int, MList[Object]]())
-    val messageIds = messageIdTracker.getOrElseUpdate(summerShards.summerIdFor(tuple._1), MList())
-    messageIds += o
+  private def add(tuple: (K, V), streamId: String, messageId: Option[AnyRef] = None): Future[Map[K, (Seq[Object], V)]] = {
+    val buffer = messageId.map { id => ListBuffer(id) }.getOrElse(Nil)
```
johnynek: for performance:

``` scala
val buffer = messageId match {
  case Some(id) => List(id)
  case None => Nil
}
```

is better than allocating the closures needed to do `.map` and `.getOrElse` if we are in a tight loop.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/collector/AggregatorOutputCollector.scala
```diff
@@ -75,43 +80,24 @@ class AggregatorOutputCollector[K, V: Semigroup](
...
+    val (k, v) = tuple
+    cacheByStreamId.getOrElseUpdate(streamId, summerBuilder.getSummer[K, (Seq[Object], V)](implicitly[Semigroup[(Seq[Object], V)]]))
```
johnynek: Seq semigroup may not be that great. Without using `sumOption` it is `O(N^2)` to aggregate.

Have you looked at `Batched`: https://github.com/twitter/algebird/blob/develop/algebird-core/src/main/scala/com/twitter/algebird/Batched.scala that deals with this?

We may want to wrap `Seq[AnyRef]` with `Batched[Seq[AnyRef]]`.

Also, can we have a private method like:

``` scala
private def getSummer =
  summerBuilder.getSummer[K, (Seq[Object], V)](implicitly[Semigroup[(Seq[Object], V)]])
```

and call that here rather than the long line?

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/collector/AggregatorOutputCollector.scala
```diff
@@ -46,27 +45,33 @@ class AggregatorOutputCollector[K, V: Semigroup](
     flushExecTimeCounter.incrBy(Time.now.inMillis - startTime.inMillis)
   }
 
-  private def convertToSummerInputFormat(flushedCache: CMap[_, _]): CMap[Int, CMap[_, _]] =
-    flushedCache.groupBy { case (k, _) => summerShards.summerIdFor(k) }
+  private def convertToSummerInputFormat(flushedCache: Map[K, (Seq[Object], V)]): TraversableOnce[OutputTuple] =
+    flushedCache.groupBy {
+      case (k, _) => summerShards.summerIdFor(k)
+    }.map {
```
johnynek: what about `.iterator.map` so we don't rebuild the `Map` after `groupBy` and `map` but only go through it once in the end?

johnynek (changes requested):

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/collector/AggregatorOutputCollector.scala
```diff
@@ -46,27 +53,33 @@ class AggregatorOutputCollector[K, V: Semigroup](
     flushExecTimeCounter.incrBy(Time.now.inMillis - startTime.inMillis)
   }
 
-  private def convertToSummerInputFormat(flushedCache: CMap[_, _]): CMap[Int, CMap[_, _]] =
-    flushedCache.groupBy { case (k, _) => summerShards.summerIdFor(k) }
+  private def convertToSummerInputFormat(flushedCache: Map[K, (OutputMessageId, V)]): TraversableOnce[OutputTuple] =
+    flushedCache.groupBy {
+      case (k, _) => summerShards.summerIdFor(k)
+    }.map {
```
johnynek: still think this should be `.iterator.map` to avoid an extra in-memory generation.

johnynek (approved):

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/collector/AggregatorOutputCollector.scala
jnievelt: Ah, sure.  It seems I missed your earlier comment about type labels so I've changed that as well.

johnynek:
:+1:

PR #697, opened 2016-10-15
Get rid of `WireFmt` and `RuntimeContext` in `OperationContainer`

ttim:
In this PR I removed `WireFmt` and `RuntimeContext` type parameters from `OperationContainer`. 

Main reason to do that was to make `OperationContainer` composable and responsible only for how to do computations on inputs. Therefore `WireFmt` moved to places which wraps `OperationContainer` in Storm.

`RuntimeContext` wasn't really used anywhere and I removed it for simplicity.

### DIFF
--- a/summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/IntermediateFlatMap.scala
+++ b/summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/IntermediateFlatMap.scala
@@ -40,5 +40,5 @@ class IntermediateFlatMap[T, U, S](
       List((List(state), Future.value(res)))
     }
 
-  override def cleanup() { lockedOp.get.close }
+  override def cleanup(): Unit = lockedOp.get.close
 }
--- a/summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/IntermediateFlatMap.scala
+++ b/summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/IntermediateFlatMap.scala
@@ -18,7 +18,6 @@ package com.twitter.summingbird.online.executor
 
 import com.twitter.util.Future
 
-import com.twitter.bijection.Injection
 import com.twitter.summingbird.online.Externalizer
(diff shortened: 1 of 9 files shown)


summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/IntermediateFlatMap.scala
```diff
@@ -46,5 +40,5 @@ class IntermediateFlatMap[T, U, S, D, RC](
       List((List(state), Future.value(res)))
     }
 
-  override def cleanup { lockedOp.get.close }
+  override def cleanup() { lockedOp.get.close }
```
johnynek: Can we avoid procedure syntax here? `(): Unit = lockedOp.get.close` will do.

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/OperationContainer.scala
```diff
@@ -17,15 +17,13 @@ limitations under the License.
 package com.twitter.summingbird.online.executor
 
 import scala.util.Try
-import com.twitter.bijection.Injection
 
-trait OperationContainer[Input, Output, State, WireFmt, RuntimeContext] {
-  def decoder: Injection[Input, WireFmt]
-  def encoder: Injection[Output, WireFmt]
+trait OperationContainer[Input, Output, State] {
+  def init() {}
```
johnynek: Let's avoid deprecated procedure syntax and below.

johnynek (changes requested):
This looks like a nice simplification. 5 type parameters is probably a hint you let things get to complex.

johnynek (approved):
:+1:

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/OperationContainer.scala
ttim: Done

jnievelt (approved):
👍

PR #699, opened 2016-11-11
Upgrade to Storm API 1.0.2

ttim:
Support modern Storm API. 

Can be merged after chill, scalding and tormenta's releases and bumping their versions.


summingbird-storm-test/src/main/scala/com/twitter/summingbird/storm/StormTestRun.scala
```diff
@@ -17,28 +17,27 @@
...
 private[storm] class MySecurityManager extends SecurityManager {
   override def checkExit(status: Int): Unit = {
-    throw new SecurityException();
+    throw new SecurityException()
   }
-  override def checkAccess(t: Thread) = {}
-  override def checkPermission(p: Permission) = {}
+  override def checkAccess(t: Thread): Unit = {}
+  override def checkPermission(p: Permission): Unit = {}
+  override def checkPermission(perm: Permission, context: scala.Any): Unit = {}
```
johnynek: please use `AnyRef` not `scala.Any` in this case. `AnyRef` is the correct corresponding type to `java.lang.Object`.

summingbird-storm-test/src/main/scala/com/twitter/summingbird/storm/StormTestRun.scala
```diff
@@ -58,22 +57,32 @@ object StormTestRun {
...
-    val oldSecManager = System.getSecurityManager()
-    System.setSecurityManager(new MySecurityManager());
+    val oldSecManager = System.getSecurityManager
+    System.setSecurityManager(new MySecurityManager())
     InflightTuples.reset()
     try {
+      Time.startSimulating()
       val cluster = new LocalCluster()
       cluster.submitTopology("test topology", plannedTopology.config, plannedTopology.topology)
-      Thread.sleep(4500)
```
johnynek: so awesome to remove these. Can you confirm there are no more `.sleep`?

summingbird-storm-test/src/main/scala/com/twitter/summingbird/storm/StormTestRun.scala
```diff
@@ -58,22 +57,32 @@ object StormTestRun {
...
+      Time.stopSimulating()
       System.setSecurityManager(oldSecManager)
     }
     require(InflightTuples.get == 0, "Inflight tuples is: %d".format(InflightTuples.get))
   }
 
+  def sleepInSimulation(millis: Int): Unit = {
+    val quantification = 10
+    0.to(millis / quantification).foreach { iter =>
+      Thread.sleep(quantification)
```
johnynek: oh no! Why do we still need to sleep?

johnynek:
what are the plans on this branch @ttim ?

ttim:
Updated this PR: now using Storm's `completeTopology` to avoid `Thread.sleep`.

@johnynek I'm planning to merge it after tormenta's release.

ttim:
@johnynek Travis is green after I disabled mima check. 
I'm unsure should I merge this change or not, because even it's green it contains two versions of kryo at the same time (I don't think it breaks anything in most cases, but still). From other point it would be a little bit more convenient for me to have this branch merged. What do you think?

johnynek:
Yeah, maybe it is good to merge this when we scalding with kryo 0.8.3

ttim:
@johnynek We released scalding so I think it's time to merge this PR (as soon as tests will pass). How does it sound for you?

After that I'm going to make one more PR (https://github.com/ttim/summingbird/tree/tabishev/bump_scalding_algebird_versions) which contains @piyushnarang changes for algebird (https://github.com/twitter/summingbird/pull/715) as well as scalding version bump.

After that as soon as @pankajroark will release new storehaus version we'll be able to do SB release, I guess with scala 2.12 support as well.

johnynek:
Yeah, I think that all sounds good.

Issue #701, opened 2016-12-21
Summingbird doesn't honor ValueCombinerCacheSize setting.


pankajroark:
This setting is not connected to anything https://github.com/twitter/summingbird/blob/8a821295d61801fb313c663a3fd8a5488c6b08fd/summingbird-storm/src/main/scala/com/twitter/summingbird/storm/BuildSummer.scala#L103. One can neither enable compactions for Async cache nor specify compaction size. Compaction can be really useful for many jobs where keeping around data for say 30 seconds can result in too much memory usage, whereas reducing this period can increase qps to KVS. Large flush period with early compactions is perfect for many use cases. 

It's actually possible to specify a custom summer but adding counters becomes problematic. Outside of the job node names are not available so different counters will need to be defined. These don't show up in the automatically generated viz charts and are thus a hassle. It's best to connect the ValueCombinerCacheSize setting as was meant to be.

PR #702, opened 2016-12-21
Support settings for enabling early compaction for async cache.

pankajroark:
Fix for #701

### DIFF
--- a/summingbird-online/src/main/scala/com/twitter/summingbird/online/OnlineDefaultConstants.scala
+++ b/summingbird-online/src/main/scala/com/twitter/summingbird/online/OnlineDefaultConstants.scala
@@ -41,6 +41,7 @@ private[summingbird] trait OnlineDefaultConstants {
   val DEFAULT_USE_ASYNC_CACHE = UseAsyncCache(false)
   val DEFAULT_ASYNC_POOL_SIZE = AsyncPoolSize(Runtime.getRuntime().availableProcessors())
   val DEFAULT_SOFT_MEMORY_FLUSH_PERCENT = SoftMemoryFlushPercent(80.0F)
+  val DEFAULT_COMPACT_VALUES = CompactValues(false)
   val DEFAULT_VALUE_COMBINER_CACHE_SIZE = ValueCombinerCacheSize(100)
   val DEFAULT_MAX_EMIT_PER_EXECUTE = MaxEmitPerExecute(Int.MaxValue)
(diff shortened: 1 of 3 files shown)


summingbird-storm/src/main/scala/com/twitter/summingbird/storm/BuildSummer.scala
```diff
@@ -118,8 +120,8 @@ object BuildSummer {
               tupleInCounter,
               tupleOutCounter,
               futurePool,
-              Compact(false),
-              CompactionSize(0))
+              Compact(doCompact.get),
```
johnynek: why not `doCompat` rather than unwrapping and re-wrapping? Same below?

pankajroark: AsyncListSum takes Compact as parameter. AsyncListSum is defined in Algebird so can't change it's signature here.

ttim (approved):

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/BuildSummer.scala
johnynek: I see, I overlooked that they are different types.

summingbird-online/src/main/scala/com/twitter/summingbird/online/OnlineDefaultConstants.scala
```diff
@@ -41,6 +41,7 @@ private[summingbird] trait OnlineDefaultConstants {
   val DEFAULT_USE_ASYNC_CACHE = UseAsyncCache(false)
   val DEFAULT_ASYNC_POOL_SIZE = AsyncPoolSize(Runtime.getRuntime().availableProcessors())
   val DEFAULT_SOFT_MEMORY_FLUSH_PERCENT = SoftMemoryFlushPercent(80.0F)
+  val DEFAULT_COMPACT_VALUES = CompactValues(false)
```
johnynek: you can't add vals to a trait without breaking binary compatibility. This trips MIMA.

Can you move this to the style of:
```
case class CompactValues(toBoolean: Boolean)
object CompactValues {
  def default: CompactValues(false)
}
```

to avoid binary fragility?

pankajroark: Sure. Done.

pankajroark:
Tests just needed a rerun, successful now. Just waiting for shipit.

johnynek:
👍

Issue #689, opened 2016-09-19
Performance of List semigroup in state aggregation is slow


pankajroark:
Summingbird online uses lists for storing state of each key in cache. 
Ultimately the value that it sums (at least in SyncSummingQueue) looks like: Map[Key, (Seq[InputState], Value)]
Thus doing plus on list(the implementation of seq that summingbird uses in Summer and FinalFlatMap) semigroup is very common and is in the critical path of summingbird online. The performance of this semigroup is very slow because we do
semigroup.plus(existingValue, newValue)
while boils down to
oldList ++ List(newValue)

oldList is typically the bigger one. So the cost of above addition is proportional to the size of bigger list. Thus if we accumulate one value at a time in the list the cost is O(n^2) rather than O(n).

I've seen this show up in YourKit profiles.

Aggregating in batch doesn't help because the top level value being summed is a Map[Key, (List[State], Value)] and GenericMapMonoid doesn't do a sumOption on underlying values in the map when doing sumOption. So ultimately plus gets called on the list of state.

One option is to change the data structure used for storing list of state, we could use a vector for example.

Another possible solution is to somehow provide a different implementation for the List semigroup used for combining state lists. Changing the order should fix it for example. List(newValue) ++ oldList would be fast and for state the order doesn't matter much. We would acknowledge newer tuples first but that shouldn't matter much, afaik nothing relies on acking order. Potentially it could result in tuples timing out but typically timeouts are pretty large for the order in a single batch to matter.

Looking to hear thoughts on this.

johnynek:
See batched:

https://github.com/twitter/algebird/blob/develop/algebird-core/src/main/scala/com/twitter/algebird/Batched.scala

Wrapping with that type will force the use of sumOption and can address the
concerns here I think (fixes the O(N^2) for sure). Try and see if that
fixes it in your kit.

pankajroark:
Batched will keep things in memory and thus increase memory footprint. A lot of the jobs are memory constrained, there is not much space available to trade for time. IMO, best would be to use a different data type for Seq[InputState], perhaps java.util.ArrayDeque[InputState] or similar that has fast append.

johnynek:
Wait, how is ArrayDeque less memory than Batched?

Do we ever actually have lazily constructed things there? I think we have
always materialized all of them anyway.

johnynek:
Isn't this the input state we are keeping? Doesn't that have to stay
materialized in memory?

pankajroark:
[...]

PR #703, opened 2016-12-24
Track state in summingbird-online as an Iterator rather than a Seq.

pankajroark:
Fix for #689 . This should avoid n^2 compute complexity when summing single element lists of Storm tuples.

### DIFF
--- a/summingbird-storm/src/main/scala/com/twitter/summingbird/storm/BaseBolt.scala
+++ b/summingbird-storm/src/main/scala/com/twitter/summingbird/storm/BaseBolt.scala
@@ -29,6 +29,7 @@ import com.twitter.summingbird.online.executor.InputState
 import com.twitter.summingbird.option.JobId
 import com.twitter.summingbird.{ Group, JobCounters, Name, SummingbirdRuntimeStats }
 import com.twitter.summingbird.online.Externalizer
+import chain.Chain
 import scala.collection.JavaConverters._
 import java.util.{ List => JList }
 import org.slf4j.{ Logger, LoggerFactory }
@@ -118,10 +119,10 @@ case class BaseBolt[I, O](jobID: JobId,
     logger.error(message, err)
   }
 
-  private def fail(inputs: Stream[InputState[Tuple]], error: Throwable): Unit = {
+  private def fail(inputs: Chain[InputState[Tuple]], error: Throwable): Unit = {
     executor.notifyFailure(inputs, error)
     if (!earlyAck) { inputs.foreach(_.fail(collector.fail(_))) }
-    logError("Storm DAG of: %d tuples failed".format(inputs.size), error)
+    logError("Storm DAG of: %d tuples failed".format(inputs.iterator.size), error)
   }
 
   override def execute(tuple: Tuple) = {
@@ -149,12 +150,13 @@ case class BaseBolt[I, O](jobID: JobId,
     }
   }
 
-  private def finish(inputs: Stream[InputState[Tuple]], results: TraversableOnce[O]) {
+  private def finish(inputs: Chain[InputState[Tuple]], results: TraversableOnce[O]) {
+    val tuples = inputs.iterator.map(_.state).toList
     var emitCount = 0
     if (hasDependants) {
       if (anchorTuples.anchor) {
         results.foreach { result =>
-          collector.emit(inputs.map(_.state).toList.asJava, encoder(result))
+          collector.emit(tuples.asJava, encoder(result))
           emitCount += 1
         }
       } else { // don't anchor
@@ -167,7 +169,7 @@ case class BaseBolt[I, O](jobID: JobId,
     // Always ack a tuple on completion:
     if (!earlyAck) { inputs.foreach(_.ack(collector.ack(_))) }
 
-    logger.debug("bolt finished processed {} linked tuples, emitted: {}", inputs.size, emitCount)
+    logger.debug("bolt finished processed {} linked tuples, emitted: {}", tuples.size, emitCount)
   }
 
   override def prepare(conf: JMap[_, _], context: TopologyContext, oc: OutputCollector) {
--- a/summingbird-storm/src/main/scala/com/twitter/summingbird/storm/BaseBolt.scala
+++ b/summingbird-storm/src/main/scala/com/twitter/summingbird/storm/BaseBolt.scala
@@ -118,7 +118,7 @@ case class BaseBolt[I, O](jobID: JobId,
     logger.error(message, err)
   }
 
-  private def fail(inputs: Seq[InputState[Tuple]], error: Throwable): Unit = {
+  private def fail(inputs: Iterator[InputState[Tuple]], error: Throwable): Unit = {
     executor.notifyFailure(inputs, error)
     if (!earlyAck) { inputs.foreach(_.fail(collector.fail(_))) }
     logError("Storm DAG of: %d tuples failed".format(inputs.size), error)
@@ -149,12 +149,12 @@ case class BaseBolt[I, O](jobID: JobId,
     }
   }
 
-  private def finish(inputs: Seq[InputState[Tuple]], results: TraversableOnce[O]) {
+  private def finish(inputs: Iterator[InputState[Tuple]], results: TraversableOnce[O]) {
     var emitCount = 0
     if (hasDependants) {
       if (anchorTuples.anchor) {
         results.foreach { result =>
-          collector.emit(inputs.map(_.state).asJava, encoder(result))
+          collector.emit(inputs.map(_.state).toList.asJava, encoder(result))
           emitCount += 1
         }
       } else { // don't anchor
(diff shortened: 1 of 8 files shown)


johnynek:
could we use `Stream` instead? the mutability of an `Iterator` makes it pretty costly to verify that it is not buggy with a code review. I really hate to have a mutable object on an API unless the performance is really necessitating it?

pankajroark:
Good idea. Changed to use stream instead of iterator.

pankajroark:
tbh Iterator is 4x faster than stream in this microbenchmark but both are way faster than List, so stream seems fine:
    val s = (0 to 10000).toList

  @Benchmark
  def listConcat(): List[Int] = {
    s ++ List(0)
  }

  @Benchmark
  def streamConcat(): Stream[Int] = {
    s.toStream ++ Stream(0)
  }

  @Benchmark
  def iterConcat(): Iterator[Int] = {
    s.toIterator ++ Iterator.single(0)
  }

Results:
[info] ToBenchmark.iterConcat    thrpt       4  85182247.256 ± 5840222.962  ops/s
[info] ToBenchmark.listConcat    thrpt        4         9937.918 ±   26498.037  ops/s
[info] ToBenchmark.streamConcat  thrpt   4  22019587.193 ± 7199199.008  ops/s

johnynek:
I would bet Batched will be as fast or faster than Iterator.

pankajroark:
I tried Batched but the code started getting complicated because of https://github.com/twitter/summingbird/blob/develop/summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/AsyncBase.scala#L46

We need to be able to pass an empty Batched here. Batched itself doesn't have a zero value it relies on the contained type. So we either have to use Option[Batched] which wouldn't be good for performance, always having to wrap/unwrap. Or we have to have the monoid of S available in this abstract class, which would require further code changes.

pankajroark:
Just for comparison iterator and Batched are indeed comparable in perf. I'd really like to use Batched if we could find a simple way or iterator otherwise. Combining input state is in the hot path.:
  val s = (0 to 10000).toList
  val bs = Batched.items(s).get
  val is = s.toIterator
  val ss = s.toStream

  @Benchmark
  def listConcat(): List[Int] = {
    s ++ List(0)
  }

  @Benchmark
  def streamConcat(): Stream[Int] = {
    ss ++ Stream(0)
  }

  @Benchmark
  def iterConcat(): Iterator[Int] = {
    is ++ Iterator.single(0)
  }

  @Benchmark
  def batchedConcat(): Batched[Int] = {
    bs.combine(Batched(0))
  }

[info] ToBenchmark.batchedConcat  thrpt    4  126300963.122 ± 46804561.489  ops/s
[info] ToBenchmark.iterConcat     thrpt         4  121149476.337 ± 46339829.598  ops/s
[info] ToBenchmark.listConcat     thrpt         4            9380.640 ±    23140.126  ops/s
[info] ToBenchmark.streamConcat   thrpt    4   24118157.945 ±  7172767.691  ops/s

pankajroark:
Any suggestions on the next steps here. I'm ok with Stream, it's still much better than List in this case.

johnynek:
I have not investigated why the tests are red.

If we can make them green with something faster. I'd be happy to do that.

I don't think using Batched would be too much work. Or we could copy this code or add this dependency:

https://github.com/non/chain/blob/master/src/main/scala/chain/Chain.scala

It is a single file library that is the more general version of Batched (it has empty), it was written by @non who also wrote Batched.

Do we really need an empty Batched? I would imagine that a `Monoid[Option[Batched[T]]]` would be almost as fast (still faster than Stream).

It is up to you. I think killing the O(N^2) is most important. Losing a constant factor of 4 is probably not a huge deal if you don't want to work on this other stuff.

Not using a mutable data structure is pretty important to me since this code has been worked on by many people now, and it is much easier to make a mistake with mutable APIs.

pankajroark:
Thanks let me try chained. I agree, constant factor of 4 is insignificant
compared to n^2. Tests are failing the Mima check.

johnynek:
Can you add the exclusions to so we can keep the tests green? It prints
what you need to add to the build to exclude those methods from erroring.

pankajroark:
Sure I'll take a look later today.

pankajroark:
Chain seems good:
[info] ToBenchmark.batchedConcat  thrpt    4  133738359.414 ±  3281958.077  ops/s
[info] ToBenchmark.chainConcat    thrpt    4      90914620.450 ± 19299667.614  ops/s
[info] ToBenchmark.iterConcat     thrpt    4       122448980.991 ± 38379171.178  ops/s
[info] ToBenchmark.streamConcat   thrpt    4     26017386.004 ±  6404231.437  ops/s

I've updated the review with now using chain.

pankajroark:
Tests pass now, just waiting for shipit to merge.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/BaseBolt.scala
```diff
@@ -149,12 +150,13 @@ case class BaseBolt[I, O](jobID: JobId,
     }
   }
 
-  private def finish(inputs: Stream[InputState[Tuple]], results: TraversableOnce[O]) {
+  private def finish(inputs: Chain[InputState[Tuple]], results: TraversableOnce[O]) {
+    val tuples = inputs.iterator.map(_.state).toList
```
johnynek: shouldn't we move this line to 158 (only materialize the `List` if we have dependants and we anchor?

johnynek (changes requested):

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/BaseBolt.scala
pankajroark: My reasoning was that we need to iterate through the chain anyway to get the size for the log statement at the end of function but I agree, materialization of list should be avoided as well in those cases. Proposed a fix.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/BaseBolt.scala
```diff
@@ -151,11 +151,13 @@ case class BaseBolt[I, O](jobID: JobId,
...
 
   private def finish(inputs: Chain[InputState[Tuple]], results: TraversableOnce[O]) {
-    val tuples = inputs.iterator.map(_.state).toList
+    var numTuples = Option.empty[Int]
     var emitCount = 0
     if (hasDependants) {
       if (anchorTuples.anchor) {
         results.foreach { result =>
+          val tuples = inputs.iterator.map(_.state).toList
+          numTuples = Some(tuples.size)
```
johnynek: list `.size` is `O(N)` not `O(1)`. So if we are going to do this, why use the var and not just use use `inputs.iterator.size`.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/BaseBolt.scala
```diff
@@ -151,11 +151,13 @@ case class BaseBolt[I, O](jobID: JobId,
   }
 
   private def finish(inputs: Chain[InputState[Tuple]], results: TraversableOnce[O]) {
-    val tuples = inputs.iterator.map(_.state).toList
+    var numTuples = Option.empty[Int]
     var emitCount = 0
     if (hasDependants) {
       if (anchorTuples.anchor) {
         results.foreach { result =>
+          val tuples = inputs.iterator.map(_.state).toList
```
johnynek: don't we want this above the foreach? We don't want to recompute `tuples` for each result. We just want it once and then reuse for all results, no?

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/BaseBolt.scala
pankajroark: Good point. We should be able to calculate the size and list in one pass though. Let me try.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/BaseBolt.scala
pankajroark: Yeah, good idea. I believe this was how it was in original code too so I didn't notice. Let me try to fix this.

johnynek:
:+1:

ttim:
What is a reason to expose `Chain` on a level of `OperationContainer` ? I'm not strictly against that but `Traversable` or even `TraversableOnce` seems more appropriate for me. 

To be more precise - `AsyncSummer` instances should be over `Chain` to avoid N^2 complexity while everything else should be over `Traversable`/`TraversableOnce`. What do you think?

johnynek:
I think making a follow up PR that minimized the scope of visibility of Chain (maybe even walking back some of the mima exclusions) would be fine. But if we have to copy to do it, I would not.

For instance, you can go: `Iterable[T] => Chain[T]` but not `Iterator[T] => Chain[T]` without a copy.

So, maybe we could use `Iterable[T]` in some cases, and internally use a `Chain[T]` to do fast concat.

That said, these are fairly "private" classes in that 99% of summingbird users would never use them. In fact, likely only storm/heron platform would use them.

ttim:
I agree.

Regarding to copying - we don't need `Iterator[T] => Chain[T]` transformation because `OperationContainer#execute` accepts single `state` element. But we need `Chain[T] => Iterable[T]` transformation which is the same (in terms of copying) as `Chain[T] => Iterator[T]`.

pankajroark:
I agree exposing chain at OperationContainer is not ideal and also that OperationContainer is used only in storm/heron platform right now, so I feel it's better to keep it simple, as is, for now. Let me create an issue to capture this though.

PR #706, opened 2017-01-11
Make sure the tail in TPNamedProducer actually names a tail

johnynek:
@pankajroark look good?


pankajroark:
:thumbsup:

johnynek:
well our terrible tests strike again.

This PR passes tests locally for me, but we are see OOM it looks like. This is such a trivial change, it if compiles, it almost has to be correct. Really hard to see how it could really cost more memory.

We need to really figure something out here.

pankajroark:
Mima is also failing btw

johnynek:
Oh yeah. Of course this takes a different type but actually the previous
type was a bug (and hopefully on one actually created one of those
themselves).

pankajroark:
Yeah, the type change looks ok.

johnynek:
our tests are so terrible.

Maybe when we upgrade storm we can finally have less flakiness.

We also need to suppress logs in the test output. Somehow the code we have for that isn't working.

PR #708, opened 2017-01-13
Clean up all but the transient warnings

johnynek:
I removed all but warnings like this:
```
[warn] /Users/oscar/oss/summingbird/summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/store/VersionedBatchStore.scala:108: no valid targets for annotation on value injection - it is discarded unused. You may specify targets with meta-annotations, e.g. @(transient @param)
[warn]   implicit @transient injection: Injection[(K2, V2), (Array[Byte], Array[Byte])], override val ordering: Ordering[K])
```

it seems like this pattern that we have been cargo-culting may not be actually doing anything, but I didn't want to take the risk (yet, but we should fix it).

All the other seemed easy to fix without doing much change to the code.

I also disabled scalariform, which is annoying in that it changes the code under you. I think we should set up scalafmt as a CI check and a command in the build to get code formatting.


summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/batch/BatchedStore.scala
```diff
@@ -108,11 +108,11 @@ trait BatchedStore[K, V] extends scalding.Store[K, V] { self =>
...
   }
 
-  protected def sumByBatches[K1, V: Semigroup](ins: TypedPipe[(Timestamp, (K1, V))],
+  protected def sumByBatches[K1, V1: Semigroup](ins: TypedPipe[(Timestamp, (K1, V1))],
     capturedBatcher: Batcher,
-    commutativity: Commutativity): TypedPipe[(LTuple2[K1, BatchID], (Timestamp, V))] = {
-    implicit val timeValueSemigroup: Semigroup[(Timestamp, V)] =
-      IteratorSums.optimizedPairSemigroup[Timestamp, V](1000)
+    commutativity: Commutativity): TypedPipe[(LTuple2[K1, BatchID], (Timestamp, V1))] = {
+    implicit val timeV1alueSemigroup: Semigroup[(Timestamp, V1)] =
```
piyushnarang: timeV1alueSemigroup, rename to timeV1Semigroup?

piyushnarang (approved):
Looks good to me but not super familiar with the SB code, so would be nice to get Pankaj's comments too.

johnynek:
Find and replace gone wrong.

summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/batch/BatchedStore.scala
pankajroark: timeValueSemigroup seems ok, the 1 got introduced by search replace I suspect.

pankajroark (approved):

johnynek:
@pankajroark good to go with the new name?

pankajroark:
👍

johnynek:
the 2.10 build keeps timing out. Not really sure what is going on. I'm hoping the #699 will just fix this.

I can't see how changing these warnings can possibly be related.

ttim:
@johnynek It's green now, I think it's better to merge this change before Storm's one.

johnynek:
@ttim maybe some issue with travis? Maybe just flakey tests.

Issue #713, opened 2017-02-09
Reducer estimation does not work with summingbird


johnynek:
In scalding, Reducer estimation runs in this code:
https://github.com/twitter/scalding/blob/72ef68b465b17fcb56a29fc13ccb8b1013430b49/scalding-core/src/main/scala/com/twitter/scalding/ExecutionContext.scala#L91

but that code is not called in summingbird:
https://github.com/twitter/summingbird/blob/5741c6570eef8292bc7956408b2a11da4e70ef88/summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/ScaldingPlatform.scala#L739

so, the options are ignored.

if we solved #710 this would probably just go away.

PR #714, opened 2017-02-10
Fix the bug with reducer estimators not working

oscar-stripe:
Addresses #713 by using the same code path scalding uses.


johnynek:
the tests are getting intolerable. We really need to look into the flakiness. I think they fail more than 1/2 the time on travis.

johnynek:
@pankajroark this should be ready to merge when you can review.

johnynek:
Maybe @benpence would also be good to look at this.

ulyssepence:
+1 @johnynek nice work! I have a branch from several months ago to do something similar, but got distracted writing the test. This should also give us the flow(step) listeners for free right (I think you spoke about this yesterday)?

johnynek:
Yes. This fixes that as well.

johnynek:
So we could remove the mutate stuff and implement that with Config flow
listeners.

pankajroark:
👍

PR #715, opened 2017-02-16
Summingbird changes for Algebird 0.13.0

piyushnarang:
Noticed that the Algebird bump broke SB in a couple of places. We might need to also bump Scalding as part of this (which I can do when we have a release out).

### DIFF
--- a/summingbird-batch-hadoop/src/test/scala/com/twitter/summingbird/batch/HDFSStateLaws.scala
+++ b/summingbird-batch-hadoop/src/test/scala/com/twitter/summingbird/batch/HDFSStateLaws.scala
@@ -16,9 +16,9 @@
 
 package com.twitter.summingbird.batch
 
-import java.util.{TimeZone, UUID}
-import com.twitter.algebird.{Intersection, Interval}
-import com.twitter.scalding.{DateParser, RichDate}
+import java.util.{ TimeZone, UUID }
+import com.twitter.algebird.{ Intersection, Interval }
+import com.twitter.scalding.{ DateParser, RichDate }
 import com.twitter.summingbird.batch.state.HDFSState
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.FileSystem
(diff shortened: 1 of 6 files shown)


piyushnarang:
@pankajroark / @johnynek / @ttim - please take a look when you can, thanks.

summingbird-batch-hadoop/src/test/scala/com/twitter/summingbird/batch/HDFSStateLaws.scala
```diff
@@ -81,7 +79,7 @@ class HDFSStateLaws extends WordSpec {
     }
   }
 
-  def leftClosedRightOpenInterval(low: Timestamp, high: Timestamp) = Interval.leftClosedRightOpen[Timestamp](low, high).right.get
+  def leftClosedRightOpenInterval(low: Timestamp, high: Timestamp) = Interval.leftClosedRightOpen[Timestamp](low, high)
```
johnynek: can we put a type on the result here? `Interval[Timestamp]`

johnynek (changes requested):

summingbird-batch-hadoop/src/test/scala/com/twitter/summingbird/batch/HDFSStateLaws.scala
piyushnarang: sure

johnynek:
👍

summingbird-batch-hadoop/src/test/scala/com/twitter/summingbird/batch/HDFSStateLaws.scala
```diff
@@ -16,15 +16,13 @@
 
 package com.twitter.summingbird.batch
 
-import java.util.{ TimeZone, UUID }
-
-import com.twitter.algebird.{ Intersection, Interval }
-import com.twitter.scalding.{ DateParser, RichDate }
+import java.util.{TimeZone, UUID}
```
ttim: nit: I think it should be with spaces (I think it's where your IDEA style settings going against our scalafmt setup).

ttim (approved):

johnynek (approved):

summingbird-batch-hadoop/src/test/scala/com/twitter/summingbird/batch/HDFSStateLaws.scala
piyushnarang: yeah seems like the Twitter defaults are set to not include the space :-).

johnynek:
looks like develop already has this. closing.

PR #716, opened 2017-03-04
Add BatchID Successible/Predecessible

johnynek:
Fix an old todo (by removing a duplicate method), and add support for Successible and Predecessible typeclasses (which are interesting for working with Bounded).

### DIFF
--- a/summingbird-batch/src/main/scala/com/twitter/summingbird/batch/BatchID.scala
+++ b/summingbird-batch/src/main/scala/com/twitter/summingbird/batch/BatchID.scala
@@ -69,17 +69,13 @@ object BatchID {
    * `[startBatch, endBatch]` (inclusive).
    */
   def range(start: BatchID, end: BatchID): Iterable[BatchID] =
-    if (start == end) List(start)
-    else if (end < start) Nil
+    if (end < start) Nil
     else {
       new Iterable[BatchID] {
-        val suc = batchIdSuccessible
-        def iterator = iterate(Option(start))(suc.next)
-          .takeWhile {
-            case Some(b) => b <= end
-            case None => false
-          }
-          .collect { case Some(b) => b }
+        def iterator = {
+          val beforeEnd = iterate(start)(_.next).takeWhile(_ < end)
+          beforeEnd ++ Iterator.single(end)
+        }
       }
     }
 
--- a/summingbird-batch/src/main/scala/com/twitter/summingbird/batch/BatchID.scala
+++ b/summingbird-batch/src/main/scala/com/twitter/summingbird/batch/BatchID.scala
@@ -19,12 +19,15 @@ package com.twitter.summingbird.batch
 import com.twitter.algebird.Monoid
 import com.twitter.algebird.{
   Empty,
-  Interval,
-  Intersection,
-  InclusiveLower,
-  ExclusiveUpper,
-  InclusiveUpper,
   ExclusiveLower,
+  ExclusiveUpper,
+  InclusiveLower,
+  InclusiveUpper,
+  Intersection,
+  Interval,
+  Monoid,
+  Successible,
+  Predecessible,
   Universe
 }
 import com.twitter.bijection.{ Bijection, Injection }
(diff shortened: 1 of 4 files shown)


johnynek:
cc @non a few of the things we hit the other day.

johnynek:
storm platform flaked. Rerunning.

johnynek:
@pankajroark can I get a review of this?

summingbird-batch/src/main/scala/com/twitter/summingbird/batch/BatchID.scala
```diff
@@ -57,8 +69,18 @@ object BatchID {
    * `[startBatch, endBatch]` (inclusive).
    */
   def range(start: BatchID, end: BatchID): Iterable[BatchID] =
-    new Iterable[BatchID] {
-      def iterator = iterate(start)(_.next).takeWhile(_ <= end)
+    if (start == end) List(start)
+    else if (end < start) Nil
+    else {
+      new Iterable[BatchID] {
```
ttim: Can we keep old way to iterate, maybe with additional check on the `end == Long.MAX_VALUE`? I think it makes code simpler.

summingbird-batch/src/main/scala/com/twitter/summingbird/batch/Batcher.scala
```diff
@@ -169,7 +169,13 @@ trait Batcher extends Serializable {
     dateToBatch(interval)(truncateUp)(truncateDown)
 
   def toInterval(b: BatchID): Interval[Timestamp] =
-    Intersection(InclusiveLower(earliestTimeOf(b)), ExclusiveUpper(earliestTimeOf(b.next)))
+    if (b.id < Long.MaxValue) {
```
ttim: Does this defense make sense? I think it makes code a little bit more noisy, while don't add value in most cases (if duration != 1ms).

summingbird-batch/src/main/scala/com/twitter/summingbird/batch/BatchID.scala
pankajroark: +1. How about:

    if (end < start) Nil
    else {
      new Iterable[BatchID] {
        def iterator = iterate(start)(_.next).takeWhile(_ < end) ++ Iterator.single(end)
      }
    }

pankajroark (changes requested):

summingbird-batch/src/main/scala/com/twitter/summingbird/batch/BatchID.scala
johnynek: I'll do @pankajroark 's approach.

I like keeping this knowledge in the Successible instance which is designed to exactly handle this task.

I don't like that `.next` may wrap around and think that API may have been a mistake at this point, and Successible does not have that problem.

summingbird-batch/src/main/scala/com/twitter/summingbird/batch/Batcher.scala
johnynek: So, the code is actually wrong if we don't do it. `b.next` will not be bigger than `b` if `b.id == Long.MaxValue`. And I don't want to have it be silently wrong.

If you prefer we can throw an error, I guess, which I don't like either.

Honestly, `earliestTimeOf(b)` can't be a total function, which is another correctness issues. It probably needs to return `Option[Timestamp]`, but that's another issue.

johnynek:
@pankajroark what do you think?

pankajroark (approved):
Lgtm

summingbird-batch/src/main/scala/com/twitter/summingbird/batch/Batcher.scala
ttim: Can we use `batchIdSuccessible.next(b)` then?
If `earliestTimeOf` will return an `Option` it would be possible to `map` `next` with `earliestTimeOf` to make this function correct.

johnynek: I'm not exactly sure what you are proposing here. We could write this code with a match:
```scala
batchIdSuccessible.next(b) match {
  case Some(bnext) => // can do left closed right open
  case None => // can't call next, so can only do inclusive
}
```
I'm not sure what `.map` has to do with it. I think you'd need a `.flatMap` actually since both `earliestTimeOf` might return Option and `next` should return `Option`.

Rather than do a big refactor of batcher and BatchID to deal with the partiality now, I just want to tackle fixing the things we can without changing the binary signature.

Note, the range bug I actually saw in a test (it would not terminate since previously it was an infinite loop).

ttim: Yes, I meant usage of `flatMap` in future, for now I'm ok with `match` and inclusive/exclusive intervals return.

johnynek:
Looks like the storm tests are almost never passing at this point. I have rerun them over and over and I don't see how this change could have broken just them on 2.11 (maybe?)

johnynek:
(they pass locally)

ttim (approved):

ttim:
This change doesn't look related to fails. I think it's safe to merge it.

PR #722, opened 2017-05-26
Bump scalacheck and scalatest versions

ttim:
This PR bumps scalacheck and scalatest version and fix all tests which started to fail.

Fixed failures:
- scalacheck generates better arbitrary functions now, this cause fail of "match scala for leftJoin jobs", "match scala for leftJoin repeated tuple leftJoin jobs" and "match scala for leftJoin with store (no dependency between the two) jobs", also it uncovered that memory law about `leftJoin` and `sumByKey` doesn't make sense
- `oneOf` generator fails in case of empty input sequence now
- some tests were failing on empty input elements, because they assumed input time interval is empty


summingbird-scalding-test/src/main/scala/com/twitter/summingbird/scalding/TestUtil.scala
```diff
@@ -90,7 +90,8 @@ object TestUtil {
    * maxTime is an exclusive upper bound.
    */
   def toTimeInterval(minTime: Long, maxTime: Long): Interval[Timestamp] =
-    Interval.leftClosedRightOpen(Timestamp(minTime), Timestamp(maxTime))
+    if (maxTime != minTime) Interval.leftClosedRightOpen(Timestamp(minTime), Timestamp(maxTime))
+    else Interval.leftClosedRightOpen(Timestamp(minTime), Timestamp(maxTime + 1))
```
johnynek: shouldn't this be empty if we have an exclusive range?

summingbird-scalding-test/src/main/scala/com/twitter/summingbird/scalding/TestUtil.scala
```diff
@@ -75,13 +75,13 @@ object TestUtil {
...
    */
-  def pruneToBatchCoveredWithTime[T](input: TraversableOnce[(Long, T)], inputRange: Interval[Timestamp], batcher: Batcher): TraversableOnce[(Long, T)] = {
+  def pruneToBatchCoveredWithTime[T](input: Traversable[(Long, T)], inputRange: Interval[Timestamp], batcher: Batcher): Traversable[(Long, T)] = {
     val batchRange = batcher.toTimestamp(batcher.batchesCoveredBy(inputRange))
     input.filter { case (ts, _) => batchRange.contains(Timestamp(ts)) }
   }
 
   /* keep just the values */
-  def pruneToBatchCovered[T](input: TraversableOnce[(Long, T)], inputRange: Interval[Timestamp], batcher: Batcher): TraversableOnce[T] = {
+  def pruneToBatchCovered[T](input: Traversable[(Long, T)], inputRange: Interval[Timestamp], batcher: Batcher): Traversable[T] = {
```
johnynek: why this change?

johnynek:
This looks fine to me.

MemoryPlatform is pretty much a toy, and for cases where there are real races in time, it makes sense that the law might fail.

summingbird-scalding-test/src/main/scala/com/twitter/summingbird/scalding/TestUtil.scala
ttim: `input` became an empty sequence in process of debugging. I'm not sure what caused this but looks like it was because something was traversable only once. I've changed this signatures and it helped. I can revert that if you want.

summingbird-scalding-test/src/main/scala/com/twitter/summingbird/scalding/TestUtil.scala
ttim: I also don't like this change. Issue is - if `input` contains zero elements we pass `minTime ==  maxTime == 0` here, have empty interval as a result, which fails after. Alternative fixes are either pass `Math.max(maxTime, 1)` (require change of all callsites) or ensure `input` contains at least one element. What do you think?

ttim:
@johnynek note that I changed `BatcherLaws` in the end. I think this change is ok because a) it shouldn't affect users, b) it makes `MillisecondBatcher` more consistent - you expect to have exactly `durationMillis` seconds in each batch, which wasn't true before (if `durationMillis` is 2, then `-1` batch will contains only `-1` millisecond).

summingbird-scalding-test/src/main/scala/com/twitter/summingbird/scalding/TestUtil.scala
ttim: Changed to `coveringTimeInterval` with more clear contract.

summingbird-scalding-test/src/main/scala/com/twitter/summingbird/scalding/TestUtil.scala
ttim: Reverted to `TraversableOnce`.

ttim:
@johnynek Do you mind to look one more time? Thanks!

summingbird-batch/src/test/scala/com/twitter/summingbird/batch/BatcherLaws.scala
```diff
@@ -101,12 +101,7 @@ object BatcherLaws extends Properties("Batcher") {
 
   val millisPerHour = 1000 * 60 * 60
 
-  def hourlyBatchFloor(batchIdx: Long): Long =
-    if (batchIdx >= 0)
-      batchIdx * millisPerHour
-    else
-      batchIdx * millisPerHour + 1
-
+  def hourlyBatchFloor(batchIdx: Long): Long = batchIdx * millisPerHour
```
johnynek, line 104: this is nicer.

johnynek (approved):
when we get the tests green, these changes look good.

johnynek:
```
[info] ! BatchID.range, toInterval and toIterable should be equivalent: Exception raised on property evaluation.
[info] > ARG_0: BatchID.9223372036854775807
[info] > ARG_1: 0
[info] > Exception: java.util.NoSuchElementException: Either.right.value on Left
```

Can you run the batcher tests locally and get them green? Might save time since travis is so slow.

ttim:
@johnynek I did run it locally and it passed. I guess test you mentioned is flaky, I'm gonna investigate it as well.

Main reason I still didn't merge is this:
> The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over).

Looks like either we started to print a little bit more logs then before (I saw size of logs exactly around 4mb in local runs) or travis changed this limit recently.
What I found is that we already try to print much less logs on travis (see last commit in regards to log4j), but it doesn't change anything. I'll merge this branch as soon as I fix this issue.

ttim:
@johnynek I tried to rerun `BatcherLaws` locally a lot of times, and I haven't seen this error. Where did you find it?

johnynek:
on the link from travis. See the red x next to pr?

ttim:
@johnynek you're right, it's a legit failure: `range, toInterval and toIterable should be equivalent` fails in case of `BatchId == MAX_LONG - 1 & diff == 0` =) It's also very rare that's why I didn't see it locally (even I ran it for 10s of times).

I'm tired of slow travis with non legit exceptions, that's why I missed this one, thanks for pointing out!

I'll fix it.

ttim:
@johnynek Actually this failure is pretty interesting. Root cause for that is - `BatchID.range` can handle intervals like [MAX_LONG, MAX_LONG], while it's impossible to handle this interval through `Interval.leftClosedRightOpen` instances. I'm wondering can we use intervals closed from both sides?

summingbird-batch/src/main/scala/com/twitter/summingbird/batch/BatchID.scala
```diff
@@ -93,7 +93,7 @@ object BatchID {
       .flatMap {
         case (min, max, cnt) =>
           if ((min + cnt) == (max + 1L)) {
-            Some(Interval.leftClosedRightOpen(min, max.next).right.get)
+            Some(Intersection(InclusiveLower(min), InclusiveUpper(max)))
```
johnynek: code *may* be assuming this interval is left closed right open.

Maybe it is better to check if `max ==BatchID(Long.MaxValue)` and only in that case do the InclusiveUpper?

ttim: Looks like some other tests started to fail with this fix, which mean we have code which *assumes* interval is left closed right open. Fix you proposed seems reasonable for me (especially for real world).

johnynek:
these tests are so frustrating.

Looks like travis killed 2.11:
```
/home/travis/.travis/job_stages: line 54:  2273 Killed                  ./sbt -Dlog4j.configuration=$LOG4J ++$TRAVIS_SCALA_VERSION clean coverage test coverageReport mimaReportBinaryIssues
```

pankajroark:
Agree, the tests are very frustrating. I'll try to spend some time on them when I get a chance.

pankajroark:
lgtm

johnynek (approved):

PR #723, opened 2017-06-01
Fix logging in storm tests

ttim:
During upgrade to Storm 1.0.x API (https://github.com/twitter/summingbird/pull/699) we broke log levels in our storm tests. There are two reasons for that:
- We didn't change package names in log4j rules for travis we use
- Storm 1.0.x implements logging through `log4j2`, and therefore we need to exclude `log4j-over-slf4j` and `log4j-slf4j-impl` ( which is kinda `slf4j-over-log4j2`) from our tests classpath. 

This PR fix both issues.


pankajroark:
lgtm

PR #726, opened 2017-06-03
Shrink jvm memory properties and use java 8

ttim:
I've limited metaspace size and shrinked code cache size to make tests memory behaviour more reliable.

### DIFF
--- a/.jvmopts
+++ b/.jvmopts
@@ -2,8 +2,8 @@
 -Xms1G
 -Xmx3G
 -Xss2m
--XX:MaxPermSize=512M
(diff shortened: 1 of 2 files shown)


ttim:
Didn't help a lot, but I think it's better to keep this properties this way.

johnynek:
Can we find a way to increase the storm timeouts?

They almost always fail now it seems.

ttim:
@johnynek I'm not sure which timeouts they are unfortunately. I think we should divide tests into three parts (scalding, storm, everything else) and investigate storm tests more, what do you think?

Also I guess it's safe to merge this PR.

Issue #725, opened 2017-06-03
Storm platform fails in case of `sumByKey` in one branch and `flatMap` in another with `flatMap`ed source fails in runtime


ttim:
Producer
```
val source = spoutSource.flatMap(e => List(e))
  source.sumByKey(store1).also(
    source.flatMap(branchFlatMap).sumByKey(store2)
  )
```
fails with `java.lang.ClassCastException: java.lang.Integer cannot be cast to scala.Tuple2` exception in runtime.

johnynek:
would be nice to fix this. Scary that something like this lasted this long.

PR #727, opened 2017-06-03
Refactor `StormLaws`

ttim:
In this PR I refactored `StormLaws` class with a way to test same `Producer` on both `Storm` and `Memory` platforms and compare their results. In future this approach can be extended to `Scalding` platform as well with a very unified way of testing.

Also I've included test for https://github.com/twitter/summingbird/issues/725, where we build `Storm` topology which fails with `ClassCastException` in runtime.


ttim:
@johnynek @pankajroark thoughts on this one?

johnynek:
👍 

Nice cleanup.

ttim:
@johnynek What I really like here is a fact we can introduce shared test suite for `Storm` and `Scalding` platforms after introduction of proper memory platform (with notion of time) and `Scalding` implementation for `CreatorCtx`.

Should I merge or better to wait for one more review from @pankajroark for example?

summingbird-storm-test/src/test/scala/com/twitter/summingbird/storm/StormTestUtils.scala
```diff
@@ -0,0 +1,183 @@
...
+    val stormSourceCreator = new StormSourceCreator()
+    val stormStoreCreator = new StormStoreCreator(seed, params)
+    val stormSinkCreator = new StormSinkCreator()
+    val stormServiceCreator = new StormServiceCreator
+    val stormCtx = new CreatorCtx(stormSourceCreator, stormStoreCreator, stormSinkCreator, stormServiceCreator)
+
+    val memory = new Memory()
+    memory.run(memory.plan(producerCreator.apply(memoryCtx)))
+
+//    assert(OnlinePlan(tail).nodes.size < 10)
```
pankajroark: remove

summingbird-storm-test/src/test/scala/com/twitter/summingbird/storm/StormTestUtils.scala
```diff
@@ -0,0 +1,183 @@
...
+  serviceCreator: ServiceCreator[P]
+) {
+  def source[T](data: List[T]): Source[P, T] = sourceCreator(data)
+  def store[K: Arbitrary, V: Arbitrary: Semigroup](id: String): P#Store[K, V] = storeCreator(id)
+  def sink[V: Ordering](id: String): P#Sink[V] = sinkCreator(id)
+  def service[K, V](fn: K => Option[V]): P#Service[K, V] = serviceCreator(fn)
+}
+
+trait SourceCreator[P <: Platform[P]] {
+  // Should be with time
```
pankajroark: Can you make this comment more clear, I'm not sure what it means in this context.

summingbird-storm-test/src/test/scala/com/twitter/summingbird/storm/StormTestUtils.scala
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  def get(): Map[String, List[_]] = sinks.mapValues(_.toList).toMap
+}
+
+object StormSinkCreator {
+  val sinks = mutable.Map[UUID, SinkContent[_]]()
+}
+
+class StormSinkCreator() extends SinkCreator[Storm] {
```
pankajroark: Comment that this is not thread safe.

summingbird-storm-test/src/test/scala/com/twitter/summingbird/storm/StormLaws.scala
```diff
@@ -173,177 +106,159 @@ class StormLaws extends WordSpec {
...
+    testStormEqualToMemory(new ProducerCreator {
+      override def apply[P <: Platform[P]](ctx: CreatorCtx[P]): TailProducer[P, Any] = {
+        val source = ctx.source(original).flatMap(e => List(e))
+        source.map(identity).sumByKey(ctx.store("store1")).also(
+          source.flatMap(branchFlatMap).sumByKey(ctx.store("store2"))
+        )
+      }
+    })
+
+//    Also fails.
```
pankajroark: When do we plan to remove these commented blocks.

summingbird-storm-test/src/test/scala/com/twitter/summingbird/storm/StormLaws.scala
```diff
@@ -99,72 +47,57 @@ object StormLaws {
...
-        TestGraphs.singleStepJob[Storm, Int, Int, Int](_, _)(testFn)
-      )
-
-    assertEquiv[Map[Int, Int]](
-      TestGraphs.singleStepInScala(original)(testFn),
-      returnedState.toScala
-    )
+    val fn = sample[Int => List[(Int, Int)]]
+
+    testStormEqualToMemory(new ProducerCreator {
```
pankajroark: Can we create a helper method to simplify this code for readability:
testStormEqualToMemory(new ProducerCreator { def apply = fn })
to
testStormEqualToMemory(fn)

pankajroark (changes requested):

summingbird-storm-test/src/test/scala/com/twitter/summingbird/storm/StormLaws.scala
johnynek: I think we should make issues about this and see if they should be gone, or if we can find a way to make them pass. I haven't looked at them much. They may be assuming a stronger contract than actually exists.

ttim: I've mentioned issue in code above (https://github.com/twitter/summingbird/issues/725), and it would be fixed soon when I'll land all my PRs. Also it's not about contract, it's just errors in our `Storm` platform implementation.

summingbird-storm-test/src/test/scala/com/twitter/summingbird/storm/StormLaws.scala
ttim: I don't know how to do this because this method has type parameter. I changed impl thou, so now it looks even better.

summingbird-storm-test/src/test/scala/com/twitter/summingbird/storm/StormTestUtils.scala
ttim: Done

summingbird-storm-test/src/test/scala/com/twitter/summingbird/storm/StormTestUtils.scala
ttim: Done

summingbird-storm-test/src/test/scala/com/twitter/summingbird/storm/StormTestUtils.scala
ttim: Done

ttim:
@johnynek @pankajroark Hi, I've reworked this change a lot with introduction of `TestPlatform` which corresponds to previous `CreatorCtx`, can you take a look? I think it's much more Summingbirdish approach.

summingbird-core-test/src/main/scala/com/twitter/summingbird/TestPlatform.scala
```diff
@@ -0,0 +1,41 @@
...
+}
+
+object TestPlatform {
+  // Test `Source` should accept data with time, not just plain data.
+  case class SourceList[T](data: List[T])
+  case class Store[K, V: Semigroup](id: String, initial: Map[K, V]) {
+    val valueSemigroup: Semigroup[V] = implicitly[Semigroup[V]]
+  }
+  case class Sink[T: Ordering](id: String) {
+    val ordering: Ordering[T] = implicitly[Ordering[T]]
```
pankajroark: Can you comment here why ordering is needed for the Test Platform? As I understand it is because one of the other platforms requires it and without this we won't be able to transform. It's not very obvious though, so a comment would be useful.

summingbird-storm-test/src/test/scala/com/twitter/summingbird/storm/StormLaws.scala
pankajroark: In that case, IMO, we can just leave a note with the commented code for now.

pankajroark (approved):
lgtm

summingbird-core-test/src/main/scala/com/twitter/summingbird/TestPlatform.scala
ttim: Removed `Ordering`. You are right - it's needed because concurrent platforms break ordering of elements. Add a comment where we check sinks content.

summingbird-storm-test/src/test/scala/com/twitter/summingbird/storm/StormLaws.scala
```diff
@@ -205,7 +205,7 @@ class StormLaws extends WordSpec {
       ))
     }
 
-//    Also fails.
+//    This is also fails, see https://github.com/twitter/summingbird/issues/725 for details.
```
pankajroark: nit: This is also -> This also

ttim: Ha, fixed

PR #728, opened 2017-06-05
Refactor `Storm` platform to introduce `Edge`s

ttim:
In order to introduce grouped `leftJoin` I would like to make tuples we send between nodes to be more precise. In particular I want to send `(K, V)` tuples as tuples with `key` and `value` fields.
 
In this PR I did a small refactoring to introduce `Edge` trait which correspond to edges in `Storm`s topology DAG - it contains a way how to serialize/deserialize data into `Fields` and how to group data over this edge.


summingbird-storm/src/main/scala/com/twitter/summingbird/storm/Edge.scala
```diff
@@ -0,0 +1,80 @@
...
+import org.apache.storm.tuple.Fields
+import java.util.{ ArrayList => JAList, List => JList }
+
+import com.twitter.summingbird.online.executor.KeyValueShards
+import org.apache.storm.topology.BoltDeclarer
+
+import scala.collection.{ Map => CMap }
+import scala.util.Try
+
+sealed trait Edge[T] {
```
pankajroark: Doc comments for trait and every public field and method.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/BaseBolt.scala
```diff
@@ -44,11 +42,10 @@ case class BaseBolt[I, O](jobID: JobId,
     metrics: () => TraversableOnce[StormMetric[_]],
     anchorTuples: AnchorTuples,
     hasDependants: Boolean,
-    outputFields: Fields,
     ackOnEntry: AckOnEntry,
     maxExecutePerSec: MaxExecutePerSecond,
-    decoder: Injection[I, JList[AnyRef]],
-    encoder: Injection[O, JList[AnyRef]],
+    inputEdges: Map[String, Edge[I]],
```
pankajroark: Add docs for all the params, I know existing code doesn't have any, but it's a good time to add them. What does the string represent. Why are there multiple incoming edges but only one outgoing edge?

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/BaseBolt.scala
```diff
@@ -192,6 +192,12 @@ case class BaseBolt[I, O](jobID: JobId,
     executor.cleanup
   }
 
+  def applyGroupings(declarer: BoltDeclarer): Unit = {
```
pankajroark: Doc comment.

pankajroark (commented):
Quick first pass is all about comments. Can you explain the reason for the change as well, i.e. how it helps with implementing grouped join.

ttim:
Currently when we send tuples over the wire we send them only in two formats: as an item (then it's just one `value` Field) or as something partially aggregated (`aggKey` and `aggValue` with `(Int, Map[K, V])` as a content). 

This breaks if we want to implement grouped join because we want to have fields we are able to group by (therefore first format is out) but also we don't want to have things partially aggregated (therefore second is out). 

Also right now we have runtime issues in topologies where we do `flatMap` and `sumByKey` in parallel branches from same `flatMap` node (see https://github.com/twitter/summingbird/issues/725). This happens because of the same reason - `flatMap` node should emit aggregated and non aggregated values at the same time.

I can see three ways to fix this issues: 

1. simplest - send all key value pairs in aggregated format, even if it's not suppose to be aggregated. 
Pros: all bolts have simple contract with one input and output format, simplest implementation. Cons: on each `(K, V)` pair we will have some size overhead (`shardId` Int, two tuple objects) and some performance overhead (but I think it's not really substantial)
2. medium - all bolts send only one type of tuples (regardless to downstream), but on receive side you can distinguish based on input component. We need second part for next use case: imagine you have `flatMap` which emits to both `sumByKey` and another `flatMap`. In this case the only way to make topology correct is to emit single element aggregated `(K, V)` on source `flatMap`, with treating them as `(K, V)` on `flatMap` downstream node and as aggregated `(K, V)` on `sumByKey` downstream node.    
Pros: contract is more or less simple, no size runtime overhead for cases which works today 
Cons: still overhead in some cases, not that simple to implement (I chase this one, that's why I have `Map[String, Edge[]]` as an input type for `Bolt`)
3. most complex and precise - implement different strategies which depends on both sending and receiving sides. 
Pros: smallest possible runtime overhead (especially if we will implement customizable `OperationContainer` which will be able to do *and* not to do partial aggregation at the same time on the same node based on downstream)
Cons: involves biggest change, especially in `BaseBolt` to support customized emits based on target `componentId`

Which one is the best, what do you think @johnynek @pankajroark ?

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/Edge.scala
```diff
@@ -0,0 +1,80 @@
...
+  val fields: Fields
+  val injection: Injection[T, JList[AnyRef]]
+  val grouping: EdgeGrouping
+}
+
+sealed trait EdgeGrouping {
+  def apply(declarer: BoltDeclarer, parentName: String): Unit
+}
```
johnynek: can we put this in a separate file: `EdgeGrouping.scala`?

And can we use the style to put inner classes to not clutter the name space:
```scala
object EdgeGrouping {
  case object Shuffle extends EdgeGrouping
  case object LocalOrShuffle extends EdgeGrouping 
}
```

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/Edge.scala
```diff
@@ -0,0 +1,80 @@
...
+case object LocalOrShuffleEdgeGrouping extends EdgeGrouping {
+  override def apply(declarer: BoltDeclarer, parentName: String): Unit =
+    declarer.localOrShuffleGrouping(parentName)
+}
+case class FieldsEdgeGrouping(fields: Fields) extends EdgeGrouping {
+  override def apply(declarer: BoltDeclarer, parentName: String): Unit =
+    declarer.fieldsGrouping(parentName, fields)
+}
+
+case class ItemEdge[T] private (edgeGrouping: EdgeGrouping) extends Edge[T] {
```
johnynek: similarly, can we put instances of `Edge` in the object:
```scala
object Edge {
  case class Item[T]...
  case class AggregatedKeyValues[K, V]...
}
```

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/FlatMapBoltProvider.scala
```diff
@@ -162,4 +156,15 @@ case class FlatMapBoltProvider(storm: Storm, jobID: JobId, stormDag: Dag[Storm],
       case None => getIntermediateFMBolt[Any, Any].asInstanceOf[BaseBolt[Any, Any]]
     }
   }
+
+  private def inputEdges[Input](): Map[String, Edge[Input]] = {
```
johnynek: scala style is to use unary methods `foo():` when there is a side effect. These don't have that so you should do:
```scala
def inputEdges[Input]: Map[String, Edge[Input]] =
```

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/Edge.scala
```diff
@@ -0,0 +1,80 @@
...
+      list.add(t.asInstanceOf[AnyRef])
+      list
+    }
+
+    override def invert(vin: JList[AnyRef]): Try[T] = Inversion.attempt(vin) { v =>
+      v.get(0).asInstanceOf[T]
+    }
+  }
+
+  def forKeyValue[K, V](): Injection[(K, V), JList[AnyRef]] = new Injection[(K, V), JList[AnyRef]] {
```
johnynek: just `def forKeyValue[K, V]: Injection...`

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/Edge.scala
```diff
@@ -0,0 +1,80 @@
...
+}
+
+case class AggregatedKeyValuesEdge[K, V](shards: KeyValueShards) extends Edge[(Int, CMap[K, V])] {
+  override val fields: Fields = new Fields("aggKey", "aggValue")
+  override val injection: Injection[(Int, CMap[K, V]), JList[AnyRef]] = EdgeInjections.forKeyValue()
+  override val grouping: EdgeGrouping = FieldsEdgeGrouping(new Fields("aggKey"))
+}
+
+object EdgeInjections {
+  def forItem[T](): Injection[T, JList[AnyRef]] = new Injection[T, JList[AnyRef]] {
```
johnynek: just `forItem[T]: Injection...` no `()`

johnynek (changes requested):

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/Edge.scala
ttim: Done, looks way better!

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/Edge.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/Edge.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/Edge.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/FlatMapBoltProvider.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/BaseBolt.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/BaseBolt.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/Edge.scala
ttim: Done

ttim:
@pankajroark added docs and explained motivation above.

I think it makes sense to use second approach for now, because I realized we need some customization of input side anyway - for `flatMap` node doing `emits` to both `flatMap` and `sumByKey` for one case we want to use shuffle grouping and for another we want to use fields grouping.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/FlatMapBoltProvider.scala
```diff
@@ -141,11 +135,11 @@ case class FlatMapBoltProvider(storm: Storm, jobID: JobId, stormDag: Dag[Storm],
       metrics.metrics,
       anchorTuples,
       stormDag.dependantsOf(node).size > 0,
-      new Fields(VALUE_FIELD),
       ackOnEntry,
       maxExecutePerSec,
-      new SingleItemInjection[ExecutorInput],
-      new SingleItemInjection[ExecutorOutput],
+      inputEdges[ExecutorInput],
+      // Output edge's grouping isn't important for now.
```
johnynek: can you comment why it odes not matter what the output grouping is?

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/FlatMapBoltProvider.scala
```diff
@@ -162,4 +156,15 @@ case class FlatMapBoltProvider(storm: Storm, jobID: JobId, stormDag: Dag[Storm],
       case None => getIntermediateFMBolt[Any, Any].asInstanceOf[BaseBolt[Any, Any]]
     }
   }
+
+  private def inputEdges[Input]: Map[String, Edge[Input]] = {
```
johnynek: seems like this should be passed in on construction from the nodes it depends on. Is that something we will get to? A win, seems to me, would be we share the same `Edge` instances between the output at one level and then input at the next, and it could reduce the chance for error there.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormPlatform.scala
```diff
@@ -205,16 +204,21 @@ abstract class Storm(options: Map[String, Options], transformConfig: Summingbird
...
         metrics.metrics,
         anchorTuples,
         shouldEmit,
-        new Fields(VALUE_FIELD),
         ackOnEntry,
         maxExecutePerSec,
-        new KeyValueInjection[Int, CMap[ExecutorKeyType, ExecutorValueType]],
-        new SingleItemInjection[ExecutorOutputType],
+        inputEdges,
+        // Output edge's grouping isn't important for now.
```
johnynek, line 220: can you comment why?

johnynek (commented):
I'm really positive about this change. Just a couple of questions for clarification.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/FlatMapBoltProvider.scala
ttim: It explained in declaration site, but what is more important - I expect that to change in subsequent review.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormPlatform.scala
ttim: Same as above.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/FlatMapBoltProvider.scala
ttim: Yes, I want to make this logic in a way you said - use same `Edge` instances for input and output with double check on edges compatibility at topology construction time (that's why I left `shardsCount` for example in grouping declaration).

pankajroark:
On strategies: the second strategy, the one you're pursuing, does seem like the right one to me. I'm still reviewing the code.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/BaseBolt.scala
```diff
@@ -27,28 +26,38 @@ import com.twitter.summingbird.{ Group, JobCounters, Name, SummingbirdRuntimeSta
...
- */
+  *  This class is used as an implementation for Storm's `Bolt`s.
+  *
+  * @param jobID is an id for current topology, used for metrics.
+  * @param metrics represents metrics we want to collect on this node.
+  * @param anchorTuples should be equal to true if you want to utilize Storm's anchoring of tuples,
+  *                     false otherwise.
+  * @param hasDependants does this node have any downstream nodes?
+  * @param ackOnEntry ack tuples in the beginning of processing.
+  * @param maxExecutePerSec limits number of executes per second, will block processing thread after.
```
pankajroark, line 43: Can we add: "Used for rate limiting."

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/BaseBolt.scala
```diff
@@ -27,28 +26,38 @@ import com.twitter.summingbird.{ Group, JobCounters, Name, SummingbirdRuntimeSta
...
+  * @param inputEdges is a map from name of downstream node to `Edge` from it.
+  * @param outputEdge is an edge from this node. To be precise there are number of output edges,
```
pankajroark: We should document this design somewhere. The package object is a common place for documenting the desing.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/BaseBolt.scala
pankajroark: Thanks, the comment looks very useful.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/Edge.scala
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    override def invert(vin: JList[AnyRef]): Try[T] = Inversion.attempt(vin) { v =>
+      v.get(0).asInstanceOf[T]
+    }
+  }
+
+  def forKeyValue[K, V]: Injection[(K, V), JList[AnyRef]] = new Injection[(K, V), JList[AnyRef]] {
+    override def apply(item: (K, V)): JAList[AnyRef] = {
+      val (key, v) = item
```
pankajroark: nit: Why key expanded and not value? Better to be consistent.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/EdgeGrouping.scala
```diff
@@ -0,0 +1,26 @@
+package com.twitter.summingbird.storm
+
+import org.apache.storm.topology.BoltDeclarer
+import org.apache.storm.tuple.{ Fields => StormFields }
+
+/**
+  * This trait is used to represent different grouping strategies in `Storm`.
+  */
+sealed trait EdgeGrouping {
+  def apply(declarer: BoltDeclarer, parentName: String): Unit
```
pankajroark: Can you comment what this does? Like what do the implementors of this trait need to do to confirm with the protocol.

pankajroark (approved):
Overall looks like a great refactoring that simplifies code. I really like that we pin down Edges and EdgeGrouping as concrete concepts.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/BaseBolt.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/BaseBolt.scala
ttim: Let's keep it until subsequent PR, there will be clearer place for this later.

pankajroark:
lgtm

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/EdgeGrouping.scala
ttim: Added

ttim:
@johnynek are you good with this change? I'm preparing next review where I did this `outputEdge` thing unnecessary.

johnynek (approved):
shipit!

PR #731, opened 2017-06-08
Introduce `Topology` class

ttim:
This PR introduces `Topology` class which corresponds to Storm's dag with nodes and edges. 

We firstly build `Topology` instance out of planned `Dag` and then build `PlannedTopology` out of it. At the moment of topology building we verify that edges in topology are compatible with each other and output format on outcoming edges is the same.


summingbird-storm/src/main/scala/com/twitter/summingbird/storm/FlatMapBoltProvider.scala
```diff
@@ -24,9 +24,12 @@ import com.twitter.summingbird.storm.option.AnchorTuples
 import com.twitter.summingbird.planner._
 import com.twitter.summingbird.online.executor
 import com.twitter.summingbird.online.FlatMapOperation
+import com.twitter.summingbird.storm.builder.Topology
 import com.twitter.summingbird.storm.planner._
 import org.slf4j.LoggerFactory
 
+import scala.collection.{ Map => CMap }
```
ulyssepence: How about `SMap` instead of `CMap`. Or even `ScalaMap`

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/Topology.scala
```diff
@@ -0,0 +1,156 @@
...
+import com.twitter.summingbird.storm.StormMetric
+import com.twitter.summingbird.storm.option.{ AckOnEntry, AnchorTuples, MaxExecutePerSecond }
+import com.twitter.tormenta.spout.{ Spout => TormentaSpout }
+import org.apache.storm.generated.StormTopology
+import org.apache.storm.topology.{ IRichBolt, TopologyBuilder }
+import org.apache.storm.{ Config => BacktypeStormConfig }
+import org.apache.storm.tuple.Tuple
+import scala.collection.{ Map => CMap }
+
+// Mark as package private?
```
ulyssepence: Mark as TODO or get rid of this

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/Topology.scala
```diff
@@ -0,0 +1,156 @@
...
+   * Set storm to tick our nodes every second to clean up finished futures
+   */
+  private def tickConfig = {
+    val boltConfig = new BacktypeStormConfig
+    boltConfig.put(BacktypeStormConfig.TOPOLOGY_TICK_TUPLE_FREQ_SECS, java.lang.Integer.valueOf(1))
+    boltConfig
+  }
+}
+
+object Topology {
```
ulyssepence: Add comments to the types in here. You are probably expecting someone to know how these things should fit together by looking at an existing platform, but I think it'd be nice to make your assumptions as clear as possible.

johnynek:
I'll look at this.

One question: is there a design doc I can look at for the planned changes.

One feature of storm we have never leveraged is that each bolt can have multiple outputs. This can be a big win for reducing serialization. I wonder if an improved design can take advantage of this.

Also, we have long wanted the ability to do key-grouped joins, and possibly merge join + mapValues + sumByKey into one node. This could reduce communtication to stores because caching might be more effective.

would love to see what is in scope for the changes rather than just polishing to the next persons satisfaction (having seen now 3 prior refactorings to suit different folks taste).

ttim:
@johnynek We tested grouped join internally and it looks very promising, so I'm making this feature production ready right now.

Basically grouped join change splits into two different things to do:
1. being able to send `(K, V)` pairs not only in case of summers but also for regular flat maps (right now SB sends either `values` tuples or partially aggregated tuples),
2. implement change to planner algorithm which creates separate flat map node with proper grouping for `leftJoin`s.

In order to implement (1) I did this refactoring to make code which builds topology clearer and simpler to fix. Also I made it more error prone, for example, it fails at construction time when you build topology defined in https://github.com/twitter/summingbird/issues/725. 

After this PR I'm going to do one more PR which implements exactly (1), and I hope it will not be complicated anymore.
And after I'm going to do one more (last) PR which implements (2) on top of this changes, and which implements grouped join (but without hot key handling for now).

/cc @benpence @pankajroark

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/Topology.scala
pankajroark: sgtm

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/Topology.scala
```diff
@@ -0,0 +1,156 @@
...
+    val boltId = Topology.BoltId[I, O](id)
+    assert(!bolts.contains(boltId))
+    (boltId, Topology(spouts, bolts.updated(boltId, bolt), edges))
+  }
+
+  def withEdge[T](edge: Topology.Edge[T]): Topology = {
+    assert(contains(edge.source) && contains(edge.dest))
+    assert(edge.source != edge.dest)
+    assert(edges.forall { anotherEdge =>
+      !(anotherEdge.source == edge.source && anotherEdge.dest == edge.dest)
```
pankajroark: may be refactor "anotherEdge.source == edge.source && anotherEdge.dest == edge.dest" into a function.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/Topology.scala
```diff
@@ -0,0 +1,156 @@
...
+    })
+
+    Topology(spouts, bolts, edges ++ List(edge))
```
pankajroark: I thought about edge :: edges but performance doesn't matter here and keeping the order may be a good idea, so lgtm.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/Topology.scala
```diff
@@ -0,0 +1,156 @@
...
+    case spoutId: Topology.SpoutId[_] => spouts.contains(spoutId)
+    case boltId: Topology.BoltId[_, _] => bolts.contains(boltId)
+  }
+
+  def incomingEdges[T](id: Topology.ReceivingId[T]): List[Topology.Edge[T]] = {
+    assert(contains(id))
+    edges.filter(_.dest == id).asInstanceOf[List[Topology.Edge[T]]]
+  }
+
+  def outcomingEdges[T](id: Topology.EmittingId[T]): List[Topology.Edge[T]] = {
```
pankajroark: outcoming -> outgoing

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/Topology.scala
```diff
@@ -0,0 +1,156 @@
...
+
+  trait EmittingId[+O] extends ComponentId
+
+  case class SpoutId[+O](override val id: String) extends EmittingId[O]
+
+  case class BoltId[-I, +O](override val id: String) extends ReceivingId[I] with EmittingId[O]
+
+  case class Edge[T](source: EmittingId[T], edgeType: EdgeType[T], dest: ReceivingId[T])
+
+  sealed trait Component {
```
pankajroark: Doc comments here and throughout.

pankajroark (commented):
First pass, only looked at Topology.scala. Continuing to review.

pankajroark:
I totally agree about shipping incrementally, groupedJoin first. We can make other optimizations after shipping groupedJoin.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
```diff
@@ -21,43 +21,62 @@ import scala.reflect.ClassTag
...
 
-  /**
-   * Set storm to tick our nodes every second to clean up finished futures
-   */
-  private def tickConfig = {
-    val boltConfig = new BacktypeStormConfig
-    boltConfig.put(BacktypeStormConfig.TOPOLOGY_TICK_TUPLE_FREQ_SECS, java.lang.Integer.valueOf(1))
-    boltConfig
-  }
+  private def scheduleNodes: Topology =
```
pankajroark: may be rename to registerNodes? I'm not sure schedule is meaningful here.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
```diff
@@ -21,43 +21,62 @@ import scala.reflect.ClassTag
...
+        topology.withBolt(getNodeName(node), scheduleSummerBolt(summerNode))._2
+      case (topology, node@(flatMapNode: FlatMapNode[Storm])) =>
+        topology.withBolt(getNodeName(node), FlatMapBoltProvider(this, flatMapNode).apply)._2
+      case (topology, node@(sourceNode: SourceNode[Storm])) =>
+        topology.withSpout(getNodeName(node), SpoutProvider(this, node).apply)._2
+    }
 
-  private def scheduleFlatMapper(node: FlatMapNode[Storm], topologyBuilder: TopologyBuilder) = {
-    val bolt: BaseBolt[Any, Any] = FlatMapBoltProvider(this, node).apply
+  private def scheduleEdges(topologyWithNode: Topology): Topology =
```
pankajroark: maybe rename to registerEdges?

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/Topology.scala
```diff
@@ -0,0 +1,156 @@
...
+
+    Topology(spouts, bolts, edges ++ List(edge))
+  }
+
+  def contains(id: Topology.ComponentId): Boolean = id match {
+    case spoutId: Topology.SpoutId[_] => spouts.contains(spoutId)
+    case boltId: Topology.BoltId[_, _] => bolts.contains(boltId)
+  }
+
+  def incomingEdges[T](id: Topology.ReceivingId[T]): List[Topology.Edge[T]] = {
```
ulyssepence, line 47: Where does this `T` come from? Do we need to lose it when creating the `Topology`?

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/Topology.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/Topology.scala
ttim: I don't think it worth it.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/Topology.scala
ttim: Yes, I think it's better to leave it as is, I changed to :+ thou.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/Topology.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/Topology.scala
ttim: I shrinked access for all new classes - to `private[summingbird]` for things which used outside and to `private[builder]` for all other things.

I added comments to most places in `Topology.scala` but not for everything, do you think we should add extensive comments? I'm not sure how stable it is, so I prefer do not add long comments for now.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/FlatMapBoltProvider.scala
ttim: We use `CMap` in other places in SB. I don't know why and I like `ScalaMap` more, but I guess it makes sense to leave it as is for now.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/Topology.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/Topology.scala
ttim: I'm not sure I understand the question. If you ask should we use `ClassTag`s then I think we should but it's not simple because we don't keep this information for now in `Producer`.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/Topology.scala
ttim: Added some comments, but I think person who will look at this code should know about `Storm` and know at least basic terminology from it.

I appreciate any points to things which confuse you at the first time seeing this code.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/FlatMapBoltProvider.scala
pankajroark: I believe CMap stands for collection.Map to disambiguate against the default collection.immutable.Map. So SMap or ScalaMap is not applicable IMO.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
```diff
@@ -21,43 +21,62 @@ import scala.reflect.ClassTag
...
-  def build(): StormTopology = {
-    val topologyBuilder = new TopologyBuilder
-    stormDag.nodes.foreach {
-      case summerNode: SummerNode[Storm] => scheduleSummerBolt(summerNode, topologyBuilder)
-      case flatMapNode: FlatMapNode[Storm] => scheduleFlatMapper(flatMapNode, topologyBuilder)
-      case sourceNode: SourceNode[Storm] => scheduleSpout(sourceNode, topologyBuilder)
-    }
-    topologyBuilder.createTopology()
-  }
+  def apply: StormTopology = scheduleEdges(scheduleNodes).build(jobId)
```
pankajroark: I feel build is a more intuitive name here, it's a very standard method name on builders to build what they are supposed to build.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
```diff
@@ -21,43 +21,62 @@ import scala.reflect.ClassTag
...
-   * Set storm to tick our nodes every second to clean up finished futures
-   */
-  private def tickConfig = {
-    val boltConfig = new BacktypeStormConfig
-    boltConfig.put(BacktypeStormConfig.TOPOLOGY_TICK_TUPLE_FREQ_SECS, java.lang.Integer.valueOf(1))
-    boltConfig
-  }
+  private def scheduleNodes: Topology =
+    stormDag.nodes.foldLeft(Topology.EMPTY) {
+      case (topology, node@(summerNode: SummerNode[Storm])) =>
```
pankajroark: Can we avoid the complicated syntax "node@(summerNode: SummerNode[Storm])". Can't we just use summerNode, why do we need node@?

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
```diff
@@ -21,43 +21,62 @@ import scala.reflect.ClassTag
...
 
-  private def scheduleFlatMapper(node: FlatMapNode[Storm], topologyBuilder: TopologyBuilder) = {
-    val bolt: BaseBolt[Any, Any] = FlatMapBoltProvider(this, node).apply
+  private def scheduleEdges(topologyWithNode: Topology): Topology =
+    stormDag.nodes.foldLeft(topologyWithNode) {
+      case (topology, node@(summerNode: SummerNode[Storm])) =>
+        val edgeType = EdgeType.AggregatedKeyValues[Any, Any](getSummerKeyValueShards(node))
+        scheduleIncomingEdges(topology, node, edgeType)
+      case (topology, node@(flatMapNode: FlatMapNode[Storm])) =>
+        val usePreferLocalDependency = getOrElse(node, DEFAULT_FM_PREFER_LOCAL_DEPENDENCY)
```
pankajroark: May be refactor portion until like 48 to separate method called flatMapEdgeType?

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/BaseBolt.scala
```diff
@@ -14,56 +14,61 @@ See the License for the specific language governing permissions and
...
+import org.apache.storm.tuple.{ Fields, Tuple, TupleImpl }
 import org.slf4j.{ Logger, LoggerFactory }
+import scala.collection.JavaConverters._
+import scala.util.{ Failure, Success }
 
 /**
   *  This class is used as an implementation for Storm's `Bolt`s.
   *
-  * @param jobID is an id for current topology, used for metrics.
+  * @param jobId is an id for current topology, used for metrics.
```
pankajroark: remove "is an": jobId id of ...
same below.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/BaseBolt.scala
```diff
@@ -14,56 +14,61 @@ See the License for the specific language governing permissions and
...
   * @param hasDependants does this node have any downstream nodes?
   * @param ackOnEntry ack tuples in the beginning of processing.
   * @param maxExecutePerSec limits number of executes per second, will block processing thread after.
   *                         Used for rate limiting.
-  * @param inputEdgeTypes is a map from name of downstream node to `Edge` from it.
-  * @param outputEdgeType is an edge from this node. To be precise there are number of output edges,
-  *                       but we expect them to have same format and we don't use their grouping
-  *                       information here, therefore it's ok to have only one instance of `Edge` here.
+  * @param inputInjections is a map from name of downstream node to format `Injection` from it.
+  * @param outputFields is Storm's `Fields` this bolt going to emit.
```
pankajroark: Storm's `Fields` that this bolt is going to emit.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/BoltBuilder.scala
```diff
@@ -0,0 +1,38 @@
+package com.twitter.summingbird.storm.builder
+
+import com.twitter.summingbird.option.JobId
+import org.apache.storm.topology.IRichBolt
+import org.apache.storm.tuple.Fields
+
+private[builder] object BoltBuilder {
```
pankajroark: Curios why you've built this wrapper by moving code out of BaseBolt. Is the intention to remove knowledge of Topology and EdgeTypes from BaseBolt? If so then may be add a comment to BaseBolt to state this design goal.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/EdgeType.scala
```diff
@@ -56,7 +54,7 @@ object EdgeType {
 }
 
 private object EdgeTypeInjections {
-  def forItem[T]: Injection[T, JList[AnyRef]] = new Injection[T, JList[AnyRef]] {
+  case class ForItem[T]() extends Injection[T, JList[AnyRef]] {
```
pankajroark: May be rename to ItemToFields? Same below.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/SpoutBuilder.scala
```diff
@@ -0,0 +1,105 @@
...
+    spout: Topology.KeyValueSpout[K, V],
+    outputEdges: List[Topology.Edge[(Int, CMap[K, V])]]
+  ): IRichSpout = {
+    val stormSpout = getStormSpout(jobId, spout.spout, spout.metrics)
+    assert(outputEdges.nonEmpty, "Should be at least one output edge from KeyValueSpout")
+    assert(
+      outputEdges.forall(edge => edge.edgeType.isInstanceOf[AggregatedKeyValues[K, V]]),
+      "All output edges from KeyValueSpout should be `AggregatedKeyValues`"
+    )
+    val shards = outputEdges.head.edgeType.asInstanceOf[AggregatedKeyValues[K, V]].shards
```
pankajroark, line 61: Assumption is there are outputEdges, which is true now but not be in future if we support Spout only topologies. May be add an assert for the assumption.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/Topology.scala
pankajroark: This could be a good utility method on Edge though, called sameEndPoints, which reads nicely: !edge.sameEndPoints(anotherEdge)

pankajroark (commented):
Second pass, almost done.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/EdgeType.scala
pankajroark: Fields is redundant actually since that's almost implied because of it being on EdgeType. I think we can just remove For and call it Item.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
ttim: Agree

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/FlatMapBoltProvider.scala
johnynek: +1 to @pankajroark

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/FlatMapBoltProvider.scala
```diff
@@ -79,10 +82,10 @@ case class FlatMapBoltProvider(builder: StormTopologyBuilder, node: StormNode) {
   private val maxEmitPerExecute = getOrElse(DEFAULT_MAX_EMIT_PER_EXECUTE)
   logger.info(s"[$nodeName] maxEmitPerExecute : ${maxEmitPerExecute.get}")
 
-  private val usePreferLocalDependency = getOrElse(DEFAULT_FM_PREFER_LOCAL_DEPENDENCY)
-  logger.info(s"[$nodeName] usePreferLocalDependency: ${usePreferLocalDependency.get}")
```
johnynek, line 83: why does this go away?

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/FlatMapBoltProvider.scala
```diff
@@ -79,10 +82,10 @@ case class FlatMapBoltProvider(builder: StormTopologyBuilder, node: StormNode) {
...
+  private val parallelism = getOrElse(DEFAULT_FM_PARALLELISM)
+  logger.info(s"[$nodeName] parallelism: ${parallelism.parHint}")
 
-  private def getFFMBolt[T, K, V](summer: SummerNode[Storm]) = {
+  private def getFFMBolt[T, K, V](summer: SummerNode[Storm]): Topology.Bolt[_, _] = {
```
johnynek: what are the `T, K, V` types here? I don't see why we need to pass type params that don't appear in the input or output, can we put them on the output?

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/BaseBolt.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/BaseBolt.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/FlatMapBoltProvider.scala
```diff
@@ -101,15 +104,12 @@ case class FlatMapBoltProvider(builder: StormTopologyBuilder, node: StormNode) {
 
     val summerBuilder = BuildSummer(builder, node)
 
-    BaseBolt(
-      builder.jobId,
+    Topology.Bolt[Input, (Int, CMap[OutputKey, OutputValue])](
```
johnynek: looks like we know the type to return?

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/BoltBuilder.scala
ttim: Reasonable, removed `BoltBuilder`.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/EdgeType.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/SpoutBuilder.scala
ttim: Not for `KeyValueSpout` I guess, we create it only in case there is summer after

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/FlatMapBoltProvider.scala
```diff
@@ -121,23 +121,19 @@ case class FlatMapBoltProvider(builder: StormTopologyBuilder, node: StormNode) {
     )
   }
 
-  def getIntermediateFMBolt[T, U] = {
+  def getIntermediateFMBolt[T, U]: Topology.Bolt[_, _] = {
```
johnynek: can't we put a return type?

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/SpoutProvider.scala
```diff
@@ -77,80 +75,55 @@ case class SpoutProvider(builder: StormTopologyBuilder, node: StormNode) {
...
   /**
    * This method creates the counters expicit for the KeyValueSpout.
    * The function to register all the counters is created with all the registered metrics.
    */
   private def getKeyValueSpout[K, V](
+    sourceParallelism: Int,
     tormentaSpout: Spout[(Timestamp, (K, V))],
-    sNode: SummerNode[Storm]): IRichSpout = {
+    sNode: SummerNode[Storm]
+  ): Topology.Spout[_] = {
```
johnynek: can't we put a type here? `Topology.Spout[(K, V)]`?

johnynek (changes requested):

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/FlatMapBoltProvider.scala
ttim: It's here now https://github.com/twitter/summingbird/pull/731/files#diff-ada7a53f54309669e9f2c41907787a5cR48

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/FlatMapBoltProvider.scala
ttim: It's too big, so I decided to skip it. Added them back.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/FlatMapBoltProvider.scala
ttim: Yep

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/FlatMapBoltProvider.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/SpoutProvider.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
```diff
@@ -80,6 +102,9 @@ case class StormTopologyBuilder(options: Map[String, Options], jobId: JobId, sto
       val maxExecutePerSec = getOrElse(node, DEFAULT_MAX_EXECUTE_PER_SEC)
       logger.info(s"[$nodeName] maxExecutePerSec : $maxExecutePerSec")
 
+      val parallelism = getOrElse(node, DEFAULT_SUMMER_PARALLELISM).parHint
```
pankajroark: shouldEmit above is unused now, may be remove that.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
```diff
@@ -21,43 +21,65 @@ import scala.reflect.ClassTag
...
+  private def getFlatMapId(node: FlatMapNode[Storm]): Topology.BoltId[_, _] =
+    Topology.BoltId(getNodeName(node))
+  private def getSourceId(node: SourceNode[Storm]): Topology.SpoutId[_] =
+    Topology.SpoutId(getNodeName(node))
+
+  private def registerSummerBolt(node: SummerNode[Storm]): Topology.Bolt[_, _] = {
     val nodeName = getNodeName(node)
 
-    def sinkBolt[K, V](summer: Summer[Storm, K, V]): BaseBolt[_, _] = {
+    def sinkBolt[K, V](summer: Summer[Storm, K, V]): Topology.Bolt[_, _] = {
```
pankajroark: Can we extract this out of this method? Very big nested methods are very cumbersome to read.

Also, may be we shouldn't call it sinkBolt unless there is a good reason to. Since we use summer terminology everywhere it will be good to be consistent.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
```diff
@@ -21,43 +21,65 @@ import scala.reflect.ClassTag
...
-    val bolt: BaseBolt[Any, Any] = FlatMapBoltProvider(this, node).apply
+  private def registerEdges(topologyWithNodes: Topology): Topology =
+    stormDag.nodes.foldLeft(topologyWithNodes) {
+      case (topology, node: SummerNode[Storm]) =>
+        val edgeType = EdgeType.AggregatedKeyValues[Any, Any](getSummerKeyValueShards(node))
+        registerIncomingEdges(topology, node, edgeType)
+      case (topology, node: FlatMapNode[Storm]) =>
+        registerIncomingEdges(topology, node, flatMapEdgeType(node))
+      case (topology, node: SourceNode[Storm]) => topology
+        // ignore, no incoming edges into sources
```
pankajroark, line 63: throw here

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
```diff
@@ -21,43 +21,65 @@ import scala.reflect.ClassTag
...
+    stormDag.dependenciesOf(node).foldLeft(topology) { (current, upstreamNode) =>
+      current.withEdge(Topology.Edge[Any](
+        getId(upstreamNode),
+        edgeType.asInstanceOf[EdgeType[Any]],
+        nodeId))
+    }
   }
 
-  private def scheduleSummerBolt(node: SummerNode[Storm], topologyBuilder: TopologyBuilder): Unit = {
+  private def getId(node: StormNode): Topology.EmittingId[_] = node match {
```
pankajroark: newline between methods

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/SpoutProvider.scala
```diff
@@ -77,80 +77,55 @@ case class SpoutProvider(builder: StormTopologyBuilder, node: StormNode) {
       case (time, (k, v)) => ((k, batcher.batchOf(time)), (time, v))
     }
     implicit val valueMonoid: Semigroup[V] = summerProducer.semigroup
-    val stormSpout = getStormSpout(formattedSummerSpout)
 
     val maxEmitPerExecute = getOrElse(node, Constants.DEFAULT_MAX_EMIT_PER_EXECUTE)
 
-    new KeyValueSpout[(K, BatchID), (Timestamp, V)](
-      stormSpout,
+    Topology.KeyValueSpout[(K, BatchID), (Timestamp, V)](
```
pankajroark: keyValueShards is unused above, remove.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/SpoutBuilder.scala
```diff
@@ -0,0 +1,105 @@
...
+import com.twitter.summingbird.{ Group, JobCounters, Name, SummingbirdRuntimeStats }
+import com.twitter.summingbird.storm._
+import com.twitter.tormenta.spout.Spout
+import org.apache.storm.metric.api.IMetric
+import org.apache.storm.task.TopologyContext
+import org.apache.storm.topology.IRichSpout
+import scala.collection.{ Map => CMap }
+
+private[builder] object SpoutBuilder {
+  def apply[T](
```
pankajroark: rename to build.

pankajroark (approved):
Minor comments. I'll let @johnynek cover the types portion. lgtm otherwise. 

You would want to run the summingbird online loadtest pretty soon to check correctness.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
ttim: I can check there is no incoming edges here, but actually it's impossible because of types. So I believe ignore is what we need.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
ttim: I did more or less on purpose %) But I think it's ok to be consistent here as well

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/SpoutProvider.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/SpoutBuilder.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
pankajroark: What's the harm in throwing if we don't expect this case to ever be hit?

ttim: Because we will hit it on each `SourceNode`, but there is nothing to do on them, because they don't have incoming edges.

ttim:
@johnynek What do you think about this change now? I specialized types where I was able to.

ttim:
@pankajroark @johnynek I did two more changes on top of what I had before:
1. extracted `SummerBoltProvider` into separate class, to be consistent with `Source` and `FlatMap` logic https://github.com/twitter/summingbird/pull/731/commits/ad1b2f9b96782748b679fcacb4ec4fedd7edc25d
2. put all common Spout/Bolt types into separate place, I believe it increased readability a lot https://github.com/twitter/summingbird/pull/731/commits/8242812d3fe9e5b1b819bb64f8ee9aae3209271f

I think this change is ready to land, what do you think?

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
```diff
@@ -12,6 +13,25 @@ import org.apache.storm.generated.StormTopology
 import org.slf4j.LoggerFactory
 import scala.reflect.ClassTag
 
+import scala.collection.{ Map => CMap }
+
+object StormTopologyBuilder {
+  type Item[T] = (Timestamp, T)
+
+  type AggregatedKey[K] = (K, BatchID)
```
pankajroark: AggregatedKey -> AggregateKey
iiuc we want to indicate key of an aggregate, rather than key that has been aggregated.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
```diff
@@ -12,6 +13,25 @@ import org.apache.storm.generated.StormTopology
...
+  type AggregatedValue[V] = (Timestamp, V)
```
pankajroark: AggregatedValue -> AggregateValue, to confirm with AggregateKey

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
```diff
@@ -12,6 +13,25 @@ import org.apache.storm.generated.StormTopology
...
+  type PartiallyAggregated[K, V] = (Int, CMap[AggregatedKey[K], AggregatedValue[V]])
```
pankajroark: PartialAggregate?
Similar convention below.

pankajroark (approved):
Minor renames suggested, lgtm otherwise.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
ttim: Agree

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
ttim: Agree

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
ttim: This case is kinda different. For key and value we want to highlight this is a type we want to aggregate over. For `PartiallyAggregated` this is a type of aggregated result, so I prefer to keep it to be `PartiallyAggregated`

pankajroark: gotcha, sgtm.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/SpoutProvider.scala
```diff
@@ -62,95 +61,77 @@ case class SpoutProvider(builder: StormTopologyBuilder, node: StormNode) {
...
+    tormentaSpout: Spout[Item[T]]
+  ): ItemSpout[T] =
+    Topology.RawSpout(sourceParallelism, spoutStormMetrics.metrics, tormentaSpout)
 
   /**
    * This is the decision logic.
    * If the spout is followed by a summer wraps the spout in KeyValueSpout
    */
-  def apply: (Int, IRichSpout) = {
+  def apply[T]: Topology.Spout[_] = {
```
johnynek, line 120: why do we have `T` here, but not in the return type?

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/BaseBolt.scala
```diff
@@ -79,6 +81,17 @@ case class BaseBolt[I, O](jobID: JobId,
   private[this] val rampPeriods = maxExecutePerSec.rampUptimeMS / PERIOD_LENGTH_MS
   private[this] val deltaPerPeriod: Long = if (rampPeriods > 0) (upperBound - lowerBound) / rampPeriods else 0
 
+  private[this] val outputFields = outputEdges.headOption.map(_.edgeType.fields).getOrElse(new Fields())
+  assert(
+    outputEdges.forall(_.edgeType.fields.toList == outputFields.toList),
+    s"$boltId: Output edges should have same `Fields` $outputEdges")
+  private[this] val outputInjection = outputEdges.headOption.map(_.edgeType.injection).orNull
```
johnynek: the `orNull` is concerning here. Why do we have that?

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/SpoutProvider.scala
ttim: This `[T]` represents type of events we emit from chain of producers we put on this `Spout`. 
For return value it can be both `ItemSpout[T]` and `AggregatedSpout[K, V]` if `T = (K, V)` (common type for them is just `Spout[_]`). 

I kept this parameter because later in code tormenta's spout has a type of `Spout[Item[T]]` regardless to what `T` is - either just `T` or `(K, V)`.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/Topology.scala
```diff
@@ -0,0 +1,188 @@
...
+   */
+  private def tickConfig = {
+    val boltConfig = new BacktypeStormConfig
+    boltConfig.put(BacktypeStormConfig.TOPOLOGY_TICK_TUPLE_FREQ_SECS, java.lang.Integer.valueOf(1))
+    boltConfig
+  }
+}
+
+private[summingbird] object Topology {
+  val EMPTY = Topology(Map(), Map(), List())
```
johnynek: can we use `val empty: Topology = ...` here rather than `EMPTY` (which is a java style, not really scala style).

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/Topology.scala
```diff
@@ -0,0 +1,188 @@
...
+
+  /**
+   * Represents id of topology's component.
+   */
+  sealed trait ComponentId {
+    val id: String
```
johnynek: can we make this a `def id: String` to avoid the risk of val init order.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/Topology.scala
```diff
@@ -0,0 +1,188 @@
...
+  case class Edge[T](source: EmittingId[T], edgeType: EdgeType[T], dest: ReceivingId[T]) {
+    def sameEndPoints(another: Edge[_]): Boolean =
+      source == another.source && dest == another.dest
+  }
+
+  /**
+   * Base trait for all components with parallelism and metrics.
+   */
+  sealed trait Component {
+    val parallelism: Int
```
johnynek: can we use `def` here? note we can override `def` with `val`.

ttim:
@johnynek What I really like about this change is fact it's making planner exceptions clear based on type signatures here https://github.com/twitter/summingbird/pull/731/files#diff-ada7a53f54309669e9f2c41907787a5cR23. From this point it's clear that planner doesn't work in case of branches with `flatMap`s & `sumByKey`s. And with two subsequent `sumByKey`s also. 

I'm preparing PR which fixes all this errors as well as introduce proper `(K, V)` handling (which enables us to do grouped join).

johnynek (commented):
some comments. I'd like to see travis go green before we merge.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/Topology.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/Topology.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/Topology.scala
ttim: Done

ttim: I didn't mark val with `override` now, is it ok?

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/BaseBolt.scala
ttim: It was for a case with empty output edges, which was handled later. I change type to `Option[Injection[[]]`.

ttim:
@johnynek Finally green build =) I guess it's safe to merge.

Interestingly one of storm tests (which do topology building for random producers) started to fail, but it's as expected (we just added more asserts) and I'm going to fix this in next PR.

johnynek (approved):
:+1:

ttim:
@pankajroark @johnynek Thank you folks for reviews.

johnynek:
👍

PR #732, opened 2017-06-09
Fix `TestPlatform` memoization

ttim:
I've discovered a bug with transformation of `Producer`s between different platforms: we don't support referential relations between created producers in `transformProducer`. This PR fixes that with memoization of created `Producer`s. 

I verified the change with running should-be-failing test locally.


pankajroark (approved):
lgtm

PR #734, opened 2017-06-15
Introduce new type of tuples we send over the network in Storm platform and fix issues with current topology builder 

ttim:
In order to add grouped join to SB online platform we need to send key value tuples over the network. Currently we send only aggregated tuples in case of summers.

This PR changes the way we build storm topology to support keyed use case and fixes known issues with current builder.

This PR:
1. Makes `Edge` trait to have two type parameters - `I` and `O`. These parameters correspond to emitted type from source side of the edge and receiving type from destination side of the edge. (https://github.com/twitter/summingbird/commit/11b94648ffbbe3ca5ce71af5dce7926a1f2c66fb).
2. Introduces `OutputFormat` which corresponds to the format of emitted values from spouts and bolts. Support for `OutputFormat` is added both to existing `BaseBolt` implementation and `KeyValueSpout`, as well as to existing `Spout` class from Tormenta (https://github.com/twitter/summingbird/commit/14fe67c68c21c0d6e156eff552d518dc30e502b2).
3. Introduces `ComponentProvider` abstraction which describes how to build a component in both aggregated and non aggregated way (https://github.com/twitter/summingbird/commit/db23a4fc159be351b61c73f4cecf2fa4c757761f).
4. Use all of above to build more robust and generic logic in `StormTopologyBuilder`.

Main points of new build logic are next:
1. each component emits values of only one type
2. `flatMap` nodes accepts tuples of type `(Timestamp, T)` where `T` corresponds to input type of `Producer`s encoded in this `flatMap` node.
3. `summer` nodes accepts tuples of type `Traversable[((K, BatchID), (Timestamp, V))]`, where `(K, V)` corresponds to type of Summer `Producer`s key and value. 
4. for non keyed values components emit `(Timestamp, T)` tuple. For keyed values it depends on the use case: if there is no keyed requirement from receiving side it's the same as for non keyed components (`(Timestamp, (K,V))`), if all downstream nodes are `Summer` nodes - then it's aggregated value as before, if some of them are non `Summer` and some of them are `Summer` - it emits `Sharded[K, V]` which is both groupable in the same way as aggregated result and can be treated as single entry for `flatMap` nodes.


summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/Summer.scala
```diff
@@ -105,11 +104,9 @@ class Summer[Key, Value: Semigroup, Event, S](
 
   override def tick = sSummer.tick.map(handleResult(_))
 
-  override def apply(state: InputState[S],
-    tupList: (Int, CMap[Key, Value])) = {
+  override def apply(state: InputState[S], innerTuples: Traversable[(Key, Value)]) = {
```
johnynek: Let's not use traversable. TraversableOnce or Iterable.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
```diff
@@ -1,102 +1,262 @@
...
 import org.apache.storm.generated.StormTopology
 import org.slf4j.LoggerFactory
 import scala.reflect.ClassTag
-
 import scala.collection.{ Map => CMap }
 
-object StormTopologyBuilder {
+/**
+ * This class contains main logic on how to build storm topology out of planned DAG.
+ * `DAG` contains three types of nodes: Source, FlatMap and Summer.
```
pankajroark: What do backticks signify? They're distracting to read. If you want to link to code then enclosing in double square brackets is the convention e.g [[com.twitter.Foo]]

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
```diff
@@ -1,102 +1,262 @@
...
+ * Built topology has next properties:
+ * 1. separate `Topology` `Component`s corresponds to separate DAG nodes
+ * 2. each `Component` emit values only of one type
+ * 3.a. `FlatMap` components accept tuples of type `Item[T]`
+ *      (and corresponds to `Producer[Storm, T]`)
+ * 3.b. `Summer` components accept tuples of type `SummerInput[K, V]`
+ *      (and corresponds to `KeyedProducer[Storm, K, V]`)
```
pankajroark: corresponds -> correspond

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/ComponentProvider.scala
```diff
@@ -0,0 +1,32 @@
+package com.twitter.summingbird.storm
+
+import com.twitter.algebird.Semigroup
+import com.twitter.summingbird.batch.Batcher
+import com.twitter.summingbird.online.executor.KeyValueShards
+import com.twitter.summingbird.storm.builder.Topology
+
+/**
+ * This trait describes how to build topology component which corresponds to `Producer`s chain.
```
pankajroark: This comment is not clear to me. What is Producer's chain?

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/Summer.scala
```diff
@@ -63,7 +62,7 @@ class Summer[Key, Value: Semigroup, Event, S](
   maxWaitingFutures: MaxWaitingFutures,
   maxWaitingTime: MaxFutureWaitTime,
   maxEmitPerExec: MaxEmitPerExecute,
-  includeSuccessHandler: IncludeSuccessHandler) extends AsyncBase[(Int, CMap[Key, Value]), Event, InputState[S]](
+  includeSuccessHandler: IncludeSuccessHandler) extends AsyncBase[Traversable[(Key, Value)], Event, InputState[S]](
```
pankajroark: Are we changing the way data flows to Summer? Instead of getting (Int, CMap[Key, Value]) now it gets Traversable[(Key, Value)].

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/EdgeGrouping.scala
```diff
@@ -22,8 +24,10 @@ private[summingbird] object EdgeGrouping {
...
       declarer.localOrShuffleGrouping(parent.id)
   }
-  case class Fields(fields: StormFields) extends EdgeGrouping {
+  case class Fields(fields: List[String]) extends EdgeGrouping {
     override def apply(declarer: BoltDeclarer, parent: Topology.EmittingId[_]): Unit =
-      declarer.fieldsGrouping(parent.id, fields)
+      declarer.fieldsGrouping(parent.id, new StormFields(ListBuffer(fields: _*).asJava))
   }
+
+  def shuffle(withLocal: Boolean): EdgeGrouping = if (withLocal) LocalOrShuffle else Shuffle
```
johnynek: this seems backwards to me. `shuffle(true)` does not look like LocalOrShuffle to me.

I also don't like Boolean flags. If we are going to keep this, can we call it `def withLocal(local: Boolean)` or someling.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/FormattedScheme.scala
```diff
@@ -0,0 +1,27 @@
...
+    try {
+      encode(decode(bytes))
+    } catch {
+      case t: Throwable => encode(handle(t))
+    }
+
+  private def encode(values: TraversableOnce[T]): JIterable[JList[AnyRef]] =
+    values.map(format.injection.apply).toIterable.asJava
+
+  override def decode(buf: ByteBuffer): TraversableOnce[T] = scheme.decode(buf)
```
johnynek, line 26: seems weird that `decode` uses `scheme` but we never call `scheme.deserialize`?

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/FormattedSpout.scala
```diff
@@ -0,0 +1,59 @@
...
+}
+
+private[builder] object FormattedSpout {
+  case class OutputCollector[T](in: SpoutOutputCollector, format: OutputFormat[T])
+    extends SpoutOutputCollector(in) {
+    override def emitDirect(
+      taskId: Int,
+      streamId: String,
+      tuple: JList[AnyRef],
+      messageId: scala.AnyRef
```
johnynek: don't need `scala.` here.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/OutputFormat.scala
```diff
@@ -0,0 +1,27 @@
...
+import java.util.{ List => JList }
+import org.apache.storm.tuple.Fields
+import scala.collection.JavaConverters.bufferAsJavaListConverter
+import scala.collection.mutable.ListBuffer
+
+/**
+ * Trait to specify storm's format for emitted values.
+ * Contains storm's tuple's fields and injection from `T` to storm's `Values` (which are java arrays).
+ */
+private[storm] case class OutputFormat[T](fields: List[String], injection: Injection[T, JList[AnyRef]]) {
```
johnynek: can this be `final`?

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/OutputFormat.scala
```diff
@@ -0,0 +1,27 @@
...
+  def asStormFields: Fields = new Fields(ListBuffer(fields: _*).asJava)
+}
+
+private[storm] object OutputFormat {
+  def get[T](outgoingEdges: Traversable[Topology.Edge[T, _]]): Option[OutputFormat[T]] = {
+    if (outgoingEdges.isEmpty) None else {
+      val format = outgoingEdges.head.format
```
johnynek: what about:
```scala
outgoingEdges.headOption.map { head =>
 
}
```

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/OutputFormat.scala
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Trait to specify storm's format for emitted values.
+ * Contains storm's tuple's fields and injection from `T` to storm's `Values` (which are java arrays).
+ */
+private[storm] case class OutputFormat[T](fields: List[String], injection: Injection[T, JList[AnyRef]]) {
+  def asStormFields: Fields = new Fields(ListBuffer(fields: _*).asJava)
+}
+
+private[storm] object OutputFormat {
+  def get[T](outgoingEdges: Traversable[Topology.Edge[T, _]]): Option[OutputFormat[T]] = {
```
johnynek: Traversable is not a great scala type. Can we use `TraversableOnce` or `Iterable`?

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/Topology.scala
```diff
@@ -137,15 +146,24 @@ private[summingbird] object Topology {
...
+
+    def deserialize(serialized: java.util.List[AnyRef]): Try[O] =
+      format.injection.invert(serialized).map(onReceiveTransform)
   }
 
   /**
    * Base trait for all components with parallelism and metrics.
    */
-  sealed trait Component {
+  sealed trait Component[+O] {
```
johnynek: can you add a comment what the type means?

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/Topology.scala
```diff
@@ -154,7 +172,7 @@ private[summingbird] object Topology {
    * Base trait for spouts.
    * There are two implementations: raw tormenta spout and key value spout.
    */
-  trait Spout[+O] extends Component
+  trait Spout[+O] extends Component[O]
```
johnynek: can we add `sealed` here? It must be sealed since Component is.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/FormattedSpout.scala
```diff
@@ -0,0 +1,59 @@
...
+    ): Unit = in.emitDirect(taskId, streamId, transform(tuple), messageId)
+
+    override def emit(
+      streamId: String,
+      tuple: JList[AnyRef],
+      messageId: scala.AnyRef
```
johnynek: no need for `scala.`

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/SummerBoltProvider.scala
```diff
@@ -12,14 +15,34 @@ import com.twitter.summingbird.storm.planner.StormNode
...
+  node: SummerNode[Storm]
+) extends ComponentProvider {
   import SummerBoltProvider._
 
-  def apply[K, V]: SummerBolt[K, V] = {
+  override def createSingle[T, O](fn: Item[T] => O): Topology.Component[O] = {
+    def create[K, V](fn: Item[(K, (Option[V], V))] => O): Topology.Component[O] =
+      bolt[K, V, O](fn)
+    // This is a legitimate conversion because we now that summer emits (K, (Option[V], V)).
+    create[Any, Any](fn.asInstanceOf[Item[(Any, (Option[Any], Any))] => O])
```
johnynek: why can't we call `bolt[Any, Any, O]` here directly?

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/SummerBoltProvider.scala
```diff
@@ -12,14 +15,34 @@ import com.twitter.summingbird.storm.planner.StormNode
...
+  builder: StormTopologyBuilder,
+  node: SummerNode[Storm]
+) extends ComponentProvider {
   import SummerBoltProvider._
 
-  def apply[K, V]: SummerBolt[K, V] = {
+  override def createSingle[T, O](fn: Item[T] => O): Topology.Component[O] = {
+    def create[K, V](fn: Item[(K, (Option[V], V))] => O): Topology.Component[O] =
+      bolt[K, V, O](fn)
+    // This is a legitimate conversion because we now that summer emits (K, (Option[V], V)).
```
johnynek: if we had this type on the `SummerNode` we might not need the cast.

johnynek (commented):
This generally looks like a lot of good changes, but the scope is pretty large.

It would be good to get the coverage of the diff up higher (I see 86% on the diff). If you install the codecov chrome plugin you can see untested methods:

https://github.com/codecov/browser-extension

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/Edges.scala
```diff
@@ -0,0 +1,115 @@
...
+    identity,
+    destId
+  )
+
+  def shuffleKeyValueToItem[K, V](
+    sourceId: Topology.EmittingId[KeyValue[K, V]],
+    destId: Topology.ReceivingId[Item[(K, V)]],
+    withLocal: Boolean
+  ): Topology.Edge[KeyValue[K, V], Item[(K, V)]] = Topology.Edge(
+    sourceId,
```
johnynek: this method is untested.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/Edges.scala
```diff
@@ -0,0 +1,115 @@
...
+ */
+private object EdgeFormats {
+  val shardKey = "shard"
+
+  def item[T]: OutputFormat[Item[T]] =
+    OutputFormat(List("timestamp", "value"), EdgeInjections.Pair())
+  def aggregated[K, V]: OutputFormat[Aggregated[K, V]] =
+    OutputFormat(List(shardKey, "aggregated"), EdgeInjections.Pair())
+  def keyValue[K, V]: OutputFormat[KeyValue[K, V]] =
+    OutputFormat(List("timestamp", "key", "value"), EdgeInjections.Triple())
```
johnynek: this branch is untested.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/FormattedScheme.scala
ttim: Because purpose of this class to actually override deserialize (which is `storm`s method) to support `Format`.

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/Summer.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/EdgeGrouping.scala
ttim: Removed this method in favor of inlining. I guess now it looks better.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/FormattedSpout.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/FormattedSpout.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/OutputFormat.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/OutputFormat.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/OutputFormat.scala
ttim: Looks good! Done.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/Topology.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/Topology.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/SummerBoltProvider.scala
ttim: Done. I though it would be more readable, but looks like no difference at all.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/Edges.scala
ttim: This was untested because we don't use `KeyValue` type for now (will be used by `leftJoin`). I removed from this change from the PR.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/Edges.scala
ttim: Same as above.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/SummerBoltProvider.scala
ttim: I had a version before where I had two separate methods `createSingle` (for `Producer[T]` basically) and `createKeyValue` for (`KeyedProducer[K, V]`). And Summer wasn't implementing first one for example. But I removed it because of complexity. I think this way it's just simpler and there are not a lot of unsafe casts in code as well so this one is ok.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
```diff
@@ -1,102 +1,238 @@
...
-    }
+  private def registerKeyValueComponent[K, V](
+    topology: Topology,
+    node: StormNode,
+    provider: ComponentProvider
+  ): Topology = outgoingSummersProps(node) match {
+    case Some(props) if props.allSummers =>
+      // Use Aggregated if possible.
+      provider.createAggregated[K, V](props.batcher, props.shards, props.semigroup
+        .asInstanceOf[Semigroup[V]]) match {
```
pankajroark: This asInstanceOf hanging on second line is a bit weird. Can we move this with what it's acting on? If needed we can add a temporary variable.

pankajroark (commented):
Quick first pass, have just scanned through. Continuing to review.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
```diff
@@ -1,102 +1,238 @@
...
+    sourceId: Topology.EmittingId[Item[T]]
+  ): Topology = stormDag.dependantsOf(source).foldLeft(topology) {
+    case (currentTopology, downstreamNode: FlatMapNode[Storm]) =>
+      currentTopology.withEdge(Edges.shuffleItemToItem[T](
+        sourceId,
+        getFMBoltId[T](downstreamNode),
+        withLocalGrouping(downstreamNode)
+      ))
+    case (_, downstreamNode: SummerNode[Storm]) =>
+      throw new Exception(s"Impossible to create item edge to summer node: " +
```
pankajroark: nit: may be change it to IllegalStateException.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/Edges.scala
```diff
@@ -0,0 +1,101 @@
...
+    sharded => Some((sharded._2, sharded._3)),
+    destId
+  )
+}
+
+/**
+ * This companion object contains all `OutputFormat`s used by Summingbird's storm topology.
+ */
+private object EdgeFormats {
+  val shardKey = "shard"
```
pankajroark: Upper case first letter for constant.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/Topology.scala
```diff
@@ -137,15 +146,25 @@ private[summingbird] object Topology {
   /**
    * Represents topology's edge with source and destination node's ids and edge type.
    */
-  case class Edge[T](source: EmittingId[T], edgeType: EdgeType[T], dest: ReceivingId[T]) {
-    def sameEndPoints(another: Edge[_]): Boolean =
+  case class Edge[I, O](
```
pankajroark: Doc comments for params, e.g. what does onReceiveTransform mean?

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/Topology.scala
```diff
@@ -137,15 +146,25 @@ private[summingbird] object Topology {
...
+    source: EmittingId[I],
+    format: OutputFormat[I],
+    grouping: EdgeGrouping,
+    onReceiveTransform: I => O,
```
pankajroark, line 165: Why is this needed? I'm concerned that we are making the edge more complex.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/Edges.scala
```diff
@@ -0,0 +1,101 @@
...
+  )
+
+  def groupedAggregatedToSummer[K, V](
+    sourceId: Topology.EmittingId[Aggregated[K, V]],
+    destId: Topology.ReceivingId[SummerInput[K, V]]
+  ): Topology.Edge[Aggregated[K, V], SummerInput[K, V]] = Topology.Edge(
+    sourceId,
+    EdgeFormats.aggregated[K, V],
+    EdgeGrouping.Fields(List(EdgeFormats.shardKey)),
+    _._2,
```
pankajroark: too terse, may be use explicit variable names for readability.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/Edges.scala
```diff
@@ -0,0 +1,101 @@
...
+
+  def shuffleShardedToItem[K, V](
+    sourceId: Topology.EmittingId[Sharded[K, V]],
+    destId: Topology.ReceivingId[Item[(K, V)]],
+    withLocal: Boolean
+  ): Topology.Edge[Sharded[K, V], Item[(K, V)]] = Topology.Edge(
+    sourceId,
+    EdgeFormats.sharded[K, V],
+    if (withLocal) EdgeGrouping.LocalOrShuffle else EdgeGrouping.Shuffle,
+    sharded => (sharded._3._1, (sharded._2._1, sharded._3._2)),
```
pankajroark: hard to read, instead of _3 etc deconstruct the args giving meaningful names.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/Edges.scala
```diff
@@ -0,0 +1,101 @@
...
+    withLocal: Boolean
+  ): Topology.Edge[Item[T], Item[T]] = Topology.Edge(
+    sourceId,
+    EdgeFormats.item[T],
+    if (withLocal) EdgeGrouping.LocalOrShuffle else EdgeGrouping.Shuffle,
+    identity,
+    destId
+  )
+
+  def groupedAggregatedToSummer[K, V](
```
pankajroark, line 28: Doc comments for all public methods here. We should make clear the terminology around Aggregated, Sharded and Item, perhaps by pointing to the right place to learn this use.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
```diff
@@ -1,102 +1,238 @@
...
 import scala.reflect.ClassTag
-
 import scala.collection.{ Map => CMap }
 
-object StormTopologyBuilder {
+/**
+ * This class contains main logic on how to build storm topology out of planned DAG.
+ * `DAG` contains three types of nodes: Source, FlatMap and Summer.
+ * Built topology has next properties:
+ * 1. separate `Topology` `Component`s corresponds to separate DAG nodes
```
pankajroark: I think it will be instructive to put an ascii diagram here e.g.:

Component ----Edge---> Component

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
```diff
@@ -1,102 +1,238 @@
...
     Topology.BoltId(getNodeName(node))
 
-  private def getFlatMapId(node: FlatMapNode[Storm]): Topology.BoltId[_, _] =
+  private def getFMBoltId[T](node: FlatMapNode[Storm]): FlatMapBoltId[T] =
     Topology.BoltId(getNodeName(node))
 
-  private def getSourceId(node: SourceNode[Storm]): Topology.SpoutId[_] =
-    Topology.SpoutId(getNodeName(node))
+  private def shouldEmitKeyValues(node: StormNode): Boolean =
+    stormDag.dependantsOf(node).exists(_.isInstanceOf[SummerNode[_]])
```
pankajroark, line 219: How about groupedJoin, we need to emit key values in that case as well right? Currently that will be a FlatMapNode.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/ComponentProvider.scala
```diff
@@ -0,0 +1,32 @@
...
+ */
+private[storm] trait ComponentProvider {
+  import StormTopologyBuilder._
+
+  /**
+   * Assuming this component corresponds to `Producer[Storm, T]` (which means it emits `Item[T]`)
```
pankajroark: This comment is also not clear. There is an example that is great but can we describe first what an implementor of ComponentProvider is supposed to implement in createSingle.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/ComponentProvider.scala
pankajroark: It's not clear why ComponentProvider needs to implement two methods below and what each out is required for at the conceptual level.

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/Summer.scala
ttim: Yes, now it accepts `Iterable[(Key, Value)]`, because sometime it's aggregated and sometime it's not.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/ComponentProvider.scala
ttim: Added an example

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
```diff
@@ -1,102 +1,238 @@
...
   }
 
-  private def getSummerId(node: SummerNode[Storm]): Topology.BoltId[_, _] =
+  private def registerItemEdges[T](
+    topology: Topology,
+    source: StormNode,
+    sourceId: Topology.EmittingId[Item[T]]
+  ): Topology = stormDag.dependantsOf(source).foldLeft(topology) {
+    case (currentTopology, downstreamNode: FlatMapNode[Storm]) =>
+      currentTopology.withEdge(Edges.shuffleItemToItem[T](
```
pankajroark, line 169: nit: Expression is too complex, may be:
      val groupedAggregatedEdge = Edges.groupedAggregatedToSummer[K, V](
        sourceId,
        getSummerBoltId[K, V](downstreamNode)
      )
      currentTopology.withEdge(groupedAggregatedEdge)

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
ttim: Splitted into lines.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
```diff
@@ -1,102 +1,238 @@
...
+  ): Topology = outgoingSummersProps(node) match {
+    case Some(props) if props.allSummers =>
+      // Use Aggregated if possible.
+      provider.createAggregated[K, V](props.batcher, props.shards, props.semigroup
+        .asInstanceOf[Semigroup[V]]) match {
+        case Some(component) =>
+          val (componentId, topologyWithComponent) = topology.withComponent(getNodeName(node), component)
+          registerAggregatedEdges[K, V](topologyWithComponent, node, componentId)
+        case None =>
+          // Fallback to Sharded.
```
pankajroark: When can this happen? May be add a comment here for that.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
```diff
@@ -1,102 +1,238 @@
...
-
-  type ItemSpout[T] = Topology.Spout[Item[T]]
-  type AggregatedSpout[K, V] = Topology.Spout[Aggregated[K, V]]
-
-  type ItemFMBolt[T, U] = Topology.Bolt[Item[T], Item[U]]
-  type AggregatedFMBolt[T, K, V] = Topology.Bolt[Item[T], Aggregated[K, V]]
-
-  type SummerOutput[K, V] = Item[(K, (Option[V], V))]
-  type SummerBolt[K, V] = Topology.Bolt[Aggregated[K, V], SummerOutput[K, V]]
+  type Sharded[K, V] = (Int, AggregateKey[K], AggregateValue[V])
```
pankajroark: Can we add comments for all types here for what they represent.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
ttim: They used to quote a code (class names or variables). It came from Markdown I guess.

Maybe I use it a little bit too much/wrongly. Do you think we should change them to [[]] everywhere? I'm not fan of using FQNs thou, I think it decreases readability.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
```diff
@@ -1,102 +1,238 @@
...
+    allSummers: Boolean,
+    batcher: Batcher,
+    semigroup: Semigroup[_],
+    shards: KeyValueShards
+  )
+
+  private def wrapSharded[K, V](
+    batcher: Batcher,
+    shards: KeyValueShards
+  ): (Item[(K, V)] => Sharded[K, V]) = { tuple =>
```
pankajroark: Can we destructure tuple to improve readability:
  ): (Item[(K, V)] => Sharded[K, V]) = { case(ts, (k, v)) =>
    val wrappedKey = (k, batcher.batchOf(ts))
    val wrappedValue = (ts, v)
    val shardId = shards.summerIdFor(wrappedKey)
    (shardId, wrappedKey, wrappedValue)

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
```diff
@@ -1,102 +1,238 @@
...
+  type SummerBoltId[K, V] = Topology.BoltId[SummerInput[K, V], _]
+
+  private case class OutgoingSummersProps(
+    allSummers: Boolean,
+    batcher: Batcher,
+    semigroup: Semigroup[_],
+    shards: KeyValueShards
+  )
+
+  private def wrapSharded[K, V](
```
pankajroark: nit: may be rename to itemToSharded, wrap doesn't clarify the intent for me, it took me a while to understand what wrap was referring to.

pankajroark (commented):
Second pass, focused on StormTopologyBuilder. Still reviewing.

ttim:
@johnynek I removed code related to `KeyValue`, now it's 91.22% mostly because of `FormattedScheme` and some assertions which shouldn't happen. I think it's ok, what do you think?

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/Edges.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/Edges.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
ttim: Yes, in future the other case for `groupedJoin` will be added here, for this review - only for summers.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/Edges.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
ttim: I'm not sure I understand about which expression exactly you are saying. IMO `edge` variable extraction doesn't improve readability and add a little bit of noise here. If separate variable here makes understanding of code here for you much easier - I'll change it.

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/Summer.scala
```diff
@@ -63,11 +62,12 @@ class Summer[Key, Value: Semigroup, Event, S](
...
   maxEmitPerExec: MaxEmitPerExecute,
-  includeSuccessHandler: IncludeSuccessHandler) extends AsyncBase[(Int, CMap[Key, Value]), Event, InputState[S]](
+  includeSuccessHandler: IncludeSuccessHandler
+) extends AsyncBase[Iterable[(Key, Value)], Event, InputState[S]](
   maxWaitingFutures,
   maxWaitingTime,
-  maxEmitPerExec) {
-
+  maxEmitPerExec
+) {
```
pankajroark, line 70: move to previous line

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
pankajroark: I think [[]] is better, even without FQNs, just for consistency.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
```diff
@@ -1,102 +1,254 @@
...
+    topology.build(jobId)
+  }
 
-  private def flatMapEdgeType(node: FlatMapNode[Storm]): EdgeType[_] = {
-    val usePreferLocalDependency = getOrElse(node, DEFAULT_FM_PREFER_LOCAL_DEPENDENCY)
-    logger.info(s"[${getNodeName(node)}] usePreferLocalDependency: ${usePreferLocalDependency.get}")
-    if (usePreferLocalDependency.get) {
-      EdgeType.itemWithLocalOrShuffleGrouping[Any]
+  private def register(topology: Topology, node: StormNode, provider: ComponentProvider): Topology =
+    if (!shouldEmitKeyValues(node)) {
```
pankajroark, line 108: very minor, may be handle the positive case first i.e. if (shouldEmitKeyValues(node)), negation is slightly harder to read. Again, very minor.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
pankajroark: It's very minor, ok to ignore, just wanted to offer a suggestion in case you also felt the same.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/SummerBoltProvider.scala
```diff
@@ -12,14 +15,31 @@ import com.twitter.summingbird.storm.planner.StormNode
...
-case class SummerBoltProvider(builder: StormTopologyBuilder, node: SummerNode[Storm]) {
+private[storm] case class SummerBoltProvider(
+  builder: StormTopologyBuilder,
+  node: SummerNode[Storm]
+) extends ComponentProvider {
   import SummerBoltProvider._
 
-  def apply[K, V]: SummerBolt[K, V] = {
+  override def createSingle[T, O](fn: Item[T] => O): Topology.Component[O] =
+    // This is a legitimate conversion because we now that summer emits (K, (Option[V], V)).
```
pankajroark: now -> know

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/BaseBolt.scala
```diff
@@ -82,19 +81,10 @@ private[builder] case class BaseBolt[I, O](
   private[this] val rampPeriods = maxExecutePerSec.rampUptimeMS / PERIOD_LENGTH_MS
   private[this] val deltaPerPeriod: Long = if (rampPeriods > 0) (upperBound - lowerBound) / rampPeriods else 0
 
-  private[this] val outputFields: Option[Fields] =
-    outputEdges.headOption.map(_.edgeType.fields)
-  assert(
-    outputEdges.forall(edge => outputFields.map(_.toList) == Some(edge.edgeType.fields.toList)),
-    s"$boltId: Output edges should have same `Fields` $outputEdges")
+  private[this] val inputEdgesMap: Map[String, Topology.Edge[_, I]] =
```
pankajroark, line 84: May be add a comment here as to why edge.source.id is unique so we don't end up overriding.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/EdgeGrouping.scala
```diff
@@ -22,8 +24,8 @@ private[summingbird] object EdgeGrouping {
     override def apply(declarer: BoltDeclarer, parent: Topology.EmittingId[_]): Unit =
       declarer.localOrShuffleGrouping(parent.id)
   }
-  case class Fields(fields: StormFields) extends EdgeGrouping {
+  case class Fields(fields: List[String]) extends EdgeGrouping {
     override def apply(declarer: BoltDeclarer, parent: Topology.EmittingId[_]): Unit =
```
pankajroark: Should we call this method register?

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/OutputFormat.scala
```diff
@@ -0,0 +1,25 @@
...
+import java.util.{ List => JList }
+import org.apache.storm.tuple.Fields
+import scala.collection.JavaConverters.bufferAsJavaListConverter
+import scala.collection.mutable.ListBuffer
+
+/**
+ * Trait to specify storm's format for emitted values.
+ * Contains storm's tuple's fields and injection from `T` to storm's `Values` (which are java arrays).
+ */
+private[storm] sealed case class OutputFormat[T](fields: List[String], injection: Injection[T, JList[AnyRef]]) {
```
pankajroark: Line too long, spread out.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/Topology.scala
pankajroark: We discussed this offline and seems ok to me.

pankajroark (changes requested):
Almost done. Still need to review SpoutBuilder and a few other files.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
ttim: I changed to [[]], but tbh I dislike it =( I think backticks much lighter and easier to read. But I don't mind to use [[]].

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
ttim: I'm not sure which diagram here to add exactly? Can you suggest something?

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/Edges.scala
ttim: I will describe them in `StormTopologyBuilder` where they are declared.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/Topology.scala
ttim: Done

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/Summer.scala
ttim: I like this way more. Do we have any concrete style guides on this?

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
ttim: I wanted to keep more general case first.

ttim: Actually even not general, but simpler case.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/SummerBoltProvider.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/EdgeGrouping.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/OutputFormat.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/BaseBolt.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/Edges.scala
ttim: About methods itself - I think they are selfdescriptive. Also note - all of them located in package private objects (as almost everything else I added).

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/ComponentProvider.scala
ttim: Any suggestions on how to describe this?

Basically this two methods correspond to non aggregated and aggregated use cases for `flatMap` and `spout`. I guess abstraction here is a little bit foggy unfortunately, but I didn't come up with better way to abstract our use cases.

ttim:
@johnynek Could you look again? I think I fixed all the comments you pointed to and simplified the change a little bit. Thanks!

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/Summer.scala
pankajroark: Actually it looks like we are not consistent in this in summingbird repository. So I'm fine with this.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/ComponentProvider.scala
pankajroark: Something on the lines of:
Create a component that emits events with single type i.e. events that don't have to have a key-value structure. e.g. ...

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/Edges.scala
pankajroark: Yeah, I'm fine with just pointing it e.g. "@see [[com.twitter.summingbird.storm.StormTopologyBuilder]] for definition of groupedAggregated"

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
pankajroark: sg

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/builder/BaseBolt.scala
pankajroark: on -> one

pankajroark (approved):
I added a few minor comments but looks good overall. LGTM

johnynek:
can we merge with head to get the fix for travis?

johnynek (approved):
Very happy with these changes. Nice work.

ttim:
Thanks @johnynek and @pankajroark for reviews!

PR #735, opened 2017-06-21
Make travis CI more stable

ttim:
I did some investigation and found that disabling any kind of parallelism for tests on travis greatly improves success rates for our tests. It doesn't feel like a good solution but I think it worth merging until we fix our tests properly. Right now it's almost impossible to pass storm tests.

What do you think @pankajroark @johnynek ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,11 +9,11 @@ matrix:
   include:
     - scala: 2.10.6
       jdk: oraclejdk8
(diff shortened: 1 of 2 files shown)


ttim:
Bad part about this change - tests will take more time (travis containers have 2 cpus), but good part - they will actually start to pass.

johnynek:
nice work! Thanks for looking into this.

johnynek (approved):

PR #737, opened 2017-06-26
Introduce ability to use grouping for `leftJoin`

ttim:
This PR adds `LeftJoinGrouping` options which makes flat map nodes which correspond to `LeftJoinProducer` grouped by key. This achieved by couple of changes:
1. introduction of `LeftJoinGrouping` option itself
2. support this option in `OnlinePlan` to force separate node for grouped `leftJoin` producers
3. support grouped `Item[(K, V)]` tuples in `StormTopologyBuilder`
4. tests for this in `StormLaws` and `TopologyTests`.

### DIFF
--- a/summingbird-storm-test/src/test/scala/com/twitter/summingbird/storm/StormLaws.scala
+++ b/summingbird-storm-test/src/test/scala/com/twitter/summingbird/storm/StormLaws.scala
@@ -224,7 +224,7 @@ class StormLaws extends WordSpec {
   "StormPlatform should work with grouped leftJoin" in {
     val leftJoinName = "leftJoin"
     val producer = TestPlatform.source(sample[List[Int]])
-      .map(sample[(Int) => (Int, Int)])
+      .map(sample[Int => (Int, Int)])
       .leftJoin(TestPlatform.service(v => Some(v))).name("leftJoin")
       .map(v => (v._1, v._2._1))
       .sumByKey(TestPlatform.store("store"))
--- a/summingbird-storm-test/src/test/scala/com/twitter/summingbird/storm/StormLaws.scala
+++ b/summingbird-storm-test/src/test/scala/com/twitter/summingbird/storm/StormLaws.scala
@@ -238,6 +238,12 @@ class StormLaws extends WordSpec {
     )))
   }
 
+  "StormPlatform should work with mapValues" in {
+    testProducer(TestPlatform.source(sample[List[(Int, Int)]])
+      .mapValues(identity)
+      .sumByKey(TestPlatform.store("store")))
+  }
+
   def testProducer[T](producer: TailProducer[TestPlatform, T])(implicit storm: Storm): Unit = {
     val memoryResult = MemoryTestExecutor(producer)
     val stormResult = StormTestExecutor(producer, storm)
--- a/summingbird-storm-test/src/test/scala/com/twitter/summingbird/storm/StormLaws.scala
+++ b/summingbird-storm-test/src/test/scala/com/twitter/summingbird/storm/StormLaws.scala
@@ -226,7 +226,7 @@ class StormLaws extends WordSpec {
     val producer = TestPlatform.source(sample[List[Int]])
       .map(sample[Int => (Int, Int)])
       .leftJoin(TestPlatform.service(v => Some(v))).name("leftJoin")
-      .map(v => (v._1, v._2._1))
+      .mapValues { case (v, _) => v }
       .sumByKey(TestPlatform.store("store"))
 
     // Test without grouped left join:
@@ -235,6 +235,21 @@ class StormLaws extends WordSpec {
     // And with:
     testProducer(producer)(Storm.local(Map(
       leftJoinName -> Options().set(LeftJoinGrouping.Grouped)
+    )))
+  }
+
+  "StormPlatform should work with grouped leftJoin and summer" in {
+    val source = TestPlatform.source(sample[List[Int]])
+      .map(sample[Int => (Int, Int)])
+
+    val producer1 = source
+      .leftJoin(TestPlatform.service(v => Some(v)))
+      .mapValues { case (v, _) => v }
+      .sumByKey(TestPlatform.store("store1"))
+    val producer2 = source.sumByKey(TestPlatform.store("store2"))
+
+    testProducer(producer1.also(producer2))(Storm.local(Map(
+      "DEFAULT" -> Options().set(LeftJoinGrouping.Grouped)
     )))
   }
 
--- a/summingbird-storm-test/src/test/scala/com/twitter/summingbird/storm/StormLaws.scala
+++ b/summingbird-storm-test/src/test/scala/com/twitter/summingbird/storm/StormLaws.scala
@@ -18,6 +18,7 @@ package com.twitter.summingbird.storm
 
 import com.twitter.summingbird._
 import com.twitter.summingbird.batch.Batcher
+import com.twitter.summingbird.online.option.LeftJoinGrouping
 import org.scalatest.WordSpec
 import org.scalacheck._
 
@@ -218,6 +219,23 @@ class StormLaws extends WordSpec {
         source.flatMap(branchFlatMap).sumByKey(TestPlatform.store("store2"))
       ))
     }
+  }
+
+  "StormPlatform should work with grouped leftJoin" in {
+    val leftJoinName = "leftJoin"
+    val producer = TestPlatform.source(sample[List[Int]])
+      .map(sample[(Int) => (Int, Int)])
+      .leftJoin(TestPlatform.service(v => Some(v))).name("leftJoin")
+      .map(v => (v._1, v._2._1))
+      .sumByKey(TestPlatform.store("store"))
+
+    // Test without grouped left join:
+    testProducer(producer)
+
+    // And with:
+    testProducer(producer)(Storm.local(Map(
+      leftJoinName -> Options().set(LeftJoinGrouping.Grouped)
+    )))
   }
 
   def testProducer[T](producer: TailProducer[TestPlatform, T])(implicit storm: Storm): Unit = {
--- a/summingbird-storm-test/src/test/scala/com/twitter/summingbird/storm/StormTopologyBuilderTests.scala
+++ b/summingbird-storm-test/src/test/scala/com/twitter/summingbird/storm/StormTopologyBuilderTests.scala
@@ -52,7 +52,7 @@ class StormTopologyBuilderTests extends WordSpec {
         .leftJoin(
           ReadableServiceFactory[Int, Int](() => ReadableStore.fromFn(v => Some(v)))
         ).name(leftJoinName)
-        .map { case (key, (value, optPrevValue)) => (key, value) }
+        .mapValues { case (value, _) => value }
         .sumByKey(TestStore.createStore[Int, Int]()._2)
 
     val topologyWithoutGrouping = buildTopology(producer)
--- a/summingbird-storm-test/src/test/scala/com/twitter/summingbird/storm/StormTopologyBuilderTests.scala
+++ b/summingbird-storm-test/src/test/scala/com/twitter/summingbird/storm/StormTopologyBuilderTests.scala
@@ -67,6 +67,7 @@ class StormTopologyBuilderTests extends WordSpec {
       leftJoinName -> Options().set(LeftJoinGrouping.Grouped)
     ))
     assert(topologyWithGrouping.spouts.size == 1)
+    // Can be two if we will be able to merge `leftJoin`, `mapValues` and `sumByKey`.
     assert(topologyWithGrouping.bolts.size == 3)
     assert(topologyWithGrouping.edges.map(_.grouping).exists { grouping =>
       grouping == EdgeGrouping.Fields(List(EdgeFormats.Key))
--- a/summingbird-storm-test/src/test/scala/com/twitter/summingbird/storm/StormTopologyBuilderTests.scala
+++ b/summingbird-storm-test/src/test/scala/com/twitter/summingbird/storm/StormTopologyBuilderTests.scala
@@ -1,0 +1,81 @@
+/*
+ Copyright 2013 Twitter, Inc.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ */
+
+package com.twitter.summingbird.storm
+
+import com.twitter.storehaus.ReadableStore
+import com.twitter.summingbird._
+import com.twitter.summingbird.batch.Batcher
+import com.twitter.summingbird.online.ReadableServiceFactory
+import com.twitter.summingbird.online.option._
+import com.twitter.summingbird.option.JobId
+import com.twitter.summingbird.planner.OnlinePlan
+import com.twitter.summingbird.storm.builder.{EdgeGrouping, Topology}
+import com.twitter.summingbird.storm.spout.TraversableSpout
+import org.scalacheck._
+import org.scalatest.WordSpec
+
+class StormTopologyBuilderTests extends WordSpec {
+  // This is dangerous, obviously. The Storm platform graphs tested
+  // here use the UnitBatcher, so the actual time extraction isn't
+  // needed.
+  implicit def extractor[T]: TimeExtractor[T] = TimeExtractor(_ => 0L)
+  implicit val batcher = Batcher.unit
+
+  /**
+   * The function tested below. We can't generate a function with
+   * ScalaCheck, as we need to know the number of tuples that the
+   * flatMap will produce.
+   */
+  val testFn = { i: Int => List((i -> i)) }
+
+  def sample[T: Arbitrary]: T = Arbitrary.arbitrary[T].sample.get
+
+  "Grouped leftJoin creates fields grouping" in {
+    val leftJoinName = "leftJoin"
+    val producer: TailProducer[Storm, _] =
+      Storm.source(TraversableSpout(sample[List[Int]]))
+        .flatMap(testFn)
+        .leftJoin(
+          ReadableServiceFactory[Int, Int](() => ReadableStore.fromFn(v => Some(v)))
+        ).name(leftJoinName)
+        .map { case (key, (value, optPrevValue)) => (key, value) }
+        .sumByKey(TestStore.createStore[Int, Int]()._2)
+
+    val topologyWithoutGrouping = buildTopology(producer)
+    assert(topologyWithoutGrouping.spouts.size == 1)
+    assert(topologyWithoutGrouping.bolts.size == 2)
+    assert(topologyWithoutGrouping.edges.map(_.grouping).forall { grouping =>
+      grouping == EdgeGrouping.Shuffle ||
+        grouping == EdgeGrouping.Fields(List(EdgeFormats.ShardKey))
+    })
+
+    val topologyWithGrouping = buildTopology(producer, Map[String, Options](
+      leftJoinName -> Options().set(LeftJoinGrouping.Grouped)
+    ))
+    assert(topologyWithGrouping.spouts.size == 1)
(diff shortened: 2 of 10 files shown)


ttim:
Also I tested this change internally at twitter and looks like it works.

summingbird-storm-test/src/test/scala/com/twitter/summingbird/storm/StormLaws.scala
```diff
@@ -220,6 +221,23 @@ class StormLaws extends WordSpec {
     }
   }
 
+  "StormPlatform should work with grouped leftJoin" in {
+    val leftJoinName = "leftJoin"
+    val producer = TestPlatform.source(sample[List[Int]])
+      .map(sample[(Int) => (Int, Int)])
```
pankajroark: Parens around the first Int look redundant.

summingbird-storm-test/src/test/scala/com/twitter/summingbird/storm/TopologyTests.scala
```diff
@@ -312,4 +315,39 @@ class TopologyTests extends WordSpec {
...
+
+  "Grouped leftJoin creates fields grouping" in {
+    val leftJoinName = "leftJoin"
+    val producer: TailProducer[Storm, _] =
+      Storm.source(TraversableSpout(sample[List[Int]]))
+        .flatMap(testFn)
+        .leftJoin(
+          ReadableServiceFactory[Int, Int](() => ReadableStore.fromFn(v => Some(v)))
+        ).name(leftJoinName)
+        .map(v => (v._1, v._2._1))
```
pankajroark: Hard to read, may be use deconstruction. map { case (k, v) => ...

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/Edges.scala
```diff
@@ -113,4 +139,19 @@ private object EdgeInjections {
       (input.get(0).asInstanceOf[T1], input.get(1).asInstanceOf[T2], input.get(2).asInstanceOf[T3])
     }
   }
+
+  case class KeyValueInjection[K, V]() extends Injection[Item[(K, V)], JList[AnyRef]] {
+    override def apply(tuple: Item[(K, V)]): JList[AnyRef] = {
+      val list = new JAList[AnyRef](3)
+      list.add(tuple._1.asInstanceOf[AnyRef])
```
pankajroark: Destructure for readability.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/Edges.scala
```diff
@@ -113,4 +139,19 @@ private object EdgeInjections {
...
+      list.add(tuple._2._1.asInstanceOf[AnyRef])
+      list.add(tuple._2._2.asInstanceOf[AnyRef])
+      list
+    }
+
+    override def invert(valueIn: JList[AnyRef]): Try[Item[(K, V)]] = Inversion.attempt(valueIn) { input =>
+      (input.get(0).asInstanceOf[Timestamp],
```
pankajroark: May be create temporary variables for readability and then
(ts, (k, v))
which is much more readable.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormPlatform.scala
```diff
@@ -179,10 +173,21 @@ abstract class Storm(options: Map[String, Options], transformConfig: Summingbird
       }
     }
 
-    PlannedTopology(config, StormTopologyBuilder(options, jobId, stormDag).build)
+    val topology = StormTopologyBuilder(options, jobId, stormDag).build
```
pankajroark, line 176: Nice, this reads simpler.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormPlatform.scala
```diff
@@ -179,10 +173,21 @@ abstract class Storm(options: Map[String, Options], transformConfig: Summingbird
...
+    PlannedTopology(config, topology.build(jobId))
   }
   def run(tail: TailProducer[Storm, _], jobName: String): Unit = run(plan(tail), jobName)
   def run(plannedTopology: PlannedTopology, jobName: String): Unit
+
+  private[storm] def planDag[T](tail: TailProducer[Storm, T]): Dag[Storm] = {
+    /*
+     * TODO: storm does not yet know about ValueFlatMapped, so remove it before
```
pankajroark: Can you create a jira ticket and link here?

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
```diff
@@ -144,8 +142,35 @@ private[storm] case class StormTopologyBuilder(options: Map[String, Options], jo
     case Some(props) =>
       registerShardedKeyValue[K, V](topology, node, provider, props.batcher, props.shards)
     case None =>
-      throw new IllegalStateException("Shouldn't happen, trying to register component which emits " +
-        "keyed values and don't have downstream summers.")
+      registerKeyValue[K, V](topology, node, provider)
```
pankajroark, line 145: nit: I just noticed this method is becoming hard to read due to nested deconstructors, perhaps refactor "case Some(props) if props.allSummers =>" part into a separate method.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
```diff
@@ -216,14 +243,23 @@ private[storm] case class StormTopologyBuilder(options: Map[String, Options], jo
...
+      case summerNode: SummerNode[Storm] => true
+    }
 
   private def withLocalGrouping(node: FlatMapNode[Storm]): Boolean = {
     val usePreferLocalDependency = getOrElse(node, DEFAULT_FM_PREFER_LOCAL_DEPENDENCY)
     logger.info(s"[${getNodeName(node)}] usePreferLocalDependency: ${usePreferLocalDependency.get}")
     usePreferLocalDependency.get
   }
 
+  private def isGroupedLeftJoinNode(node: FlatMapNode[Storm]): Boolean =
```
pankajroark, line 257: This doesn't need to be a method on this class. Should perhaps be a method on the companion and unit tested.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
```diff
@@ -92,16 +93,13 @@ private[storm] object StormTopologyBuilder {
...
-    val topology = stormDag.nodes.foldLeft(Topology.empty) {
-      case (currentTopology, node: SummerNode[Storm]) =>
-        register(currentTopology, node, SummerBoltProvider(this, node))
-      case (currentTopology, node: FlatMapNode[Storm]) =>
-        register(currentTopology, node, FlatMapBoltProvider(this, node))
-      case (currentTopology, node: SourceNode[Storm]) =>
-        register(currentTopology, node, SpoutProvider(this, node))
-    }
-    topology.build(jobId)
+  def build: Topology = stormDag.nodes.foldLeft(Topology.empty) {
```
pankajroark, line 96: We should add unit tests for StormTopologyBuilder to cover this feature.

pankajroark (changes requested):
Looks great overall. I think good to ship after adding unit tests for StormTopologyBuilder.

summingbird-storm-test/src/test/scala/com/twitter/summingbird/storm/StormLaws.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/Edges.scala
ttim: Done

summingbird-storm-test/src/test/scala/com/twitter/summingbird/storm/TopologyTests.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/Edges.scala
ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormPlatform.scala
ttim: Actually this is not needed anymore, removed this code and added test for `mapValues` into `StormLaws`. Thanks!

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
ttim: I added test to `TopologyTests`, but I think separate `StormTopologyBuilderTests` would be nicer.

ttim: Done

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
ttim: Unfortunately in this method we access `Options` which makes both extraction and testing much harder. I'm agree this is not the best place for this method but IMO it's best place now.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
ttim: This method represents high level strategy how to deal with key value components, all further logic extracted into `registerAggregatedEdges`, `registerShardedKeyValue` and `registerKeyValue`. I can extract `aggregated` part further but I prefer to keep it as is. Also I doubt it will grow more.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormPlatform.scala
```diff
@@ -162,13 +162,7 @@ abstract class Storm(options: Map[String, Options], transformConfig: Summingbird
   def withConfigUpdater(fn: SummingbirdConfig => SummingbirdConfig): Storm
 
   def plan[T](tail: TailProducer[Storm, T]): PlannedTopology = {
-    /*
-     * TODO: storm does not yet know about ValueFlatMapped, so remove it before
-     * planning
-     */
-    val dagOptimizer = new DagOptimizer[Storm] {}
-    val stormTail = dagOptimizer.optimize(tail, dagOptimizer.ValueFlatMapToFlatMap)
```
johnynek, line 170: this was already done? We have support for ValueFlatMapped now? If so, can we combine `leftJoin.mapValues.sumByKey` into a single node if `leftJoin` is grouped?

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
```diff
@@ -216,14 +243,23 @@ private[storm] case class StormTopologyBuilder(options: Map[String, Options], jo
...
+    }
 
   private def withLocalGrouping(node: FlatMapNode[Storm]): Boolean = {
     val usePreferLocalDependency = getOrElse(node, DEFAULT_FM_PREFER_LOCAL_DEPENDENCY)
     logger.info(s"[${getNodeName(node)}] usePreferLocalDependency: ${usePreferLocalDependency.get}")
     usePreferLocalDependency.get
   }
 
+  private def isGroupedLeftJoinNode(node: FlatMapNode[Storm]): Boolean =
+    node.members.last.isInstanceOf[LeftJoinedProducer[Storm, _, _, _]] &&
```
johnynek: I'm suspect of the `.last` here. This means we can't add a `mapValues` or `flatMapValues` on the end.

Why don't we want `node.members.exists { case _: LeftJoinedProducer[_, _, _, _] => true; case _ => false } && `?

summingbird-storm-test/src/test/scala/com/twitter/summingbird/storm/StormTopologyBuilderTests.scala
```diff
@@ -0,0 +1,81 @@
...
+
+  "Grouped leftJoin creates fields grouping" in {
+    val leftJoinName = "leftJoin"
+    val producer: TailProducer[Storm, _] =
+      Storm.source(TraversableSpout(sample[List[Int]]))
+        .flatMap(testFn)
+        .leftJoin(
+          ReadableServiceFactory[Int, Int](() => ReadableStore.fromFn(v => Some(v)))
+        ).name(leftJoinName)
+        .map { case (key, (value, optPrevValue)) => (key, value) }
```
johnynek: can you write this with `.mapValues { case (value, _) => value }`, which has the potential for optimization in scalding at least, so people can see examples where `mapValues/collectValues/flatMapValues` is preferred?

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/Edges.scala
```diff
@@ -113,4 +139,20 @@ private object EdgeInjections {
...
+      case (timestamp, (key, value)) =>
+        val list = new JAList[AnyRef](3)
+        list.add(timestamp.asInstanceOf[AnyRef])
+        list.add(key.asInstanceOf[AnyRef])
+        list.add(value.asInstanceOf[AnyRef])
+        list
+    }
+
+    override def invert(valueIn: JList[AnyRef]): Try[Item[(K, V)]] = Inversion.attempt(valueIn) { input =>
+      val (timestamp, key, value) = (input.get(0), input.get(1), input.get(2))
```
pankajroark: I'm not sure scalac will optimize this out, so there could be performance penalty. May be just split to multiple lines. 
val timestamp = input.get(0).asInstanceOf[Timestamp]
...

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
pankajroark: sg

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
pankajroark: sg, since extraction is harder it's ok.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
pankajroark: It would also be good to add tests with mapValues and flatMapValues to capture the edge cases that @johnynek pointed out.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormPlatform.scala
ttim: We have support on the level we can build `flatMap` operation over it (https://github.com/twitter/summingbird/commit/44fb5c9#diff-1d4313fb460f1921fc349a8eb2ee902fR85). But it's not fully supported in a way to affect `OnlinePlan`. I would like to do this in future, but I guess it's out of this PR. It's about when do not trigger new node creation, and this one is when to trigger it =)

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
ttim: @johnynek `.last` is here refers to first element in producers chain unfortunately (I added a comment on `.members` in previous PR).

I don't think we need tests on this edge cases because we don't support this edge cases yet.

summingbird-storm-test/src/test/scala/com/twitter/summingbird/storm/StormTopologyBuilderTests.scala
ttim: Sure. Looks better now, thanks!

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/Edges.scala
ttim: I thought about this but decided it should be optimized by compiler =) changed to your way.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
ttim: Added a comment in test that number of bolts can be `-1` in case of support of `mapValues`.

ttim: Also I added `firstProducer` method on `Dag`s `Node`, to make it look better.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormPlatform.scala
pankajroark: I agree, better to go incrementally. We can do that optimization in separate review.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
```diff
@@ -307,7 +307,7 @@ private[storm] case class StormTopologyBuilder(options: Map[String, Options], jo
     }
 
   private[storm] def get[T <: AnyRef: ClassTag](node: StormNode): Option[(String, T)] = {
-    val producer = node.members.last
+    val producer = node.firstProducer.get
```
johnynek: any reason not to do:
```scala
node.firstProducer.flatMap { producer =>
  Options.getFirst[T](options, stormDag.producerToPriorityNames(producer))
}
```
here?

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/Edges.scala
```diff
@@ -48,6 +60,17 @@ private[storm] object Edges {
     destId
   )
 
+  def groupedShardedToItem[K, V](
+    sourceId: Topology.EmittingId[Sharded[K, V]],
+    destId: Topology.ReceivingId[Item[(K, V)]]
+  ): Topology.Edge[Sharded[K, V], Item[(K, V)]] = Topology.Edge(
```
johnynek, line 66: this function is untested. Any way to get some test coverage of it?

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormPlatform.scala
johnynek: wait, I'm confused why you removed this code then.

Can you link to where in the storm platform we handle `flatMapValues`? The link you have above points to the optimizer rule you are removing here.

pankajroark (approved):
LGTM

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormPlatform.scala
ttim: @johnynek because I pointed to the wrong code, sorry %) Here is a right link https://github.com/twitter/summingbird/commit/44fb5c9#diff-81c566a323153de58aa7f47b84ee941aR54.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
ttim: Changed to `.flatMap` way.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/Edges.scala
ttim: Sure, done!

johnynek:
👍 

100% test coverage of this diff. Nice!

johnynek:
It will be exciting when we can merge `leftJoin.mapValues.sumByKey` into one node to reduce serialization.

I bet this is actually a pretty common pattern: look-up by the key, do some transformation, aggregate a value.

pankajroark:
Agree with @johnynek most of our realtime topologies have sumByKey and leftJoin when present is almost always a preparatory step for it, so leftJoin.mapValues.sumByKey is pretty common when leftJoin is used. That will be a very useful optimization.

Issue #711, opened 2017-01-27
Make it easy to add metrics to SummerConstructor


pankajroark:
There are couple things that make adding metrics for replacement SummerBuilder very difficult. 

1. The metrics for regular summer have the node name in them, e.g: summingbird.Tail/tuplesIn. The overridden summer has to know the nodename to use the same metric names.
2. An implicit job id is required for creating summingbird counters. Most jobs don't set job id by default and are forced to create one for this case.

Instead Counter should be passed to SummerBuilder. SummerBuilder is seldom used so one possibility is to accept breaking backward compatibility in this case. Otherwise a new option can be defined and this one deprecated.

PR #738, opened 2017-06-27
Refactor `SummerBuilder` to make it possible to use `Counters` on `SummerBuilder` creation

ttim:
Currently the way to customize `Summer` instances is not perfect - we have an old 'legacy' way based on a lot of different options and 'new' way with setting `SummerBuilder` directly. 
At the same time with new way it's impossible to fully utilize `Counters` because `jobId` is needed to create them. Also new way is quite verbose.

This PR adds a way to use `Counters` in `SummerBuilder` and couple of `SummerBuilder`s:  `Null`, `sync` and `async`.

### DIFF
--- a/summingbird-online/src/main/scala/com/twitter/summingbird/online/option/AllOpts.scala
+++ b/summingbird-online/src/main/scala/com/twitter/summingbird/online/option/AllOpts.scala
@@ -2,7 +2,8 @@ package com.twitter.summingbird.online.option
 
 import com.twitter.util.Duration
 import com.twitter.algebird.Semigroup
-import com.twitter.algebird.util.summer.AsyncSummer
+import com.twitter.algebird.util.summer.{ AsyncSummer, Incrementor }
+import com.twitter.summingbird.{ Counter, Name }
 
 case class OnlineSuccessHandler(handlerFn: Unit => Unit)
 
@@ -98,8 +99,17 @@ trait SummerBuilder extends Serializable {
 
 /**
  * The SummerConstructor option, set this instead of CacheSize, AsyncPoolSize, etc.. to provide how to construct the aggregation for this bolt
+ * @see [[Summers]] for useful [[SummerConstructor]]s.
  */
-case class SummerConstructor(get: SummerBuilder)
+case class SummerConstructor(get: SummerConstructor.Context => SummerBuilder)
+
+object SummerConstructor {
+  def apply(get: SummerBuilder): SummerConstructor = SummerConstructor(_ => get)
+
+  trait Context {
+    def counter(name: Name): Counter with Incrementor
+  }
+}
 
 /**
  * How many instances/tasks of this flatmap task should be spawned in the environment
--- a/summingbird-online/src/main/scala/com/twitter/summingbird/online/option/AllOpts.scala
+++ b/summingbird-online/src/main/scala/com/twitter/summingbird/online/option/AllOpts.scala
@@ -99,20 +99,20 @@ trait SummerBuilder extends Serializable {
 
 /**
  * The SummerConstructor option, set this instead of CacheSize, AsyncPoolSize, etc.. to provide how to construct the aggregation for this bolt
- * @see [[Summers]] for useful [[SummerConstructor]]s.
+ * @see [[Summers]] for useful [[SummerWithCountersBuilder]]s.
  */
-case class SummerConstructor(get: SummerConstructorSpec)
+case class SummerConstructor(get: SummerWithCountersBuilder)
 
-trait SummerConstructorSpec {
-  def builder(counter: Name => Counter with Incrementor): SummerBuilder
+trait SummerWithCountersBuilder {
+  def create(counter: Name => Counter with Incrementor): SummerBuilder
 }
 
 object SummerConstructor {
   def apply(get: SummerBuilder): SummerConstructor =
     SummerConstructor(DeprecatedSummerConstructorSpec(get))
 
-  private case class DeprecatedSummerConstructorSpec(get: SummerBuilder) extends SummerConstructorSpec {
-    override def builder(counter: (Name) => Counter with Incrementor): SummerBuilder = get
+  private case class DeprecatedSummerConstructorSpec(get: SummerBuilder) extends SummerWithCountersBuilder {
+    override def create(counter: (Name) => Counter with Incrementor): SummerBuilder = get
   }
 }
 
--- a/summingbird-online/src/main/scala/com/twitter/summingbird/online/option/AllOpts.scala
+++ b/summingbird-online/src/main/scala/com/twitter/summingbird/online/option/AllOpts.scala
@@ -101,13 +101,18 @@ trait SummerBuilder extends Serializable {
  * The SummerConstructor option, set this instead of CacheSize, AsyncPoolSize, etc.. to provide how to construct the aggregation for this bolt
  * @see [[Summers]] for useful [[SummerConstructor]]s.
  */
-case class SummerConstructor(get: SummerConstructor.Context => SummerBuilder)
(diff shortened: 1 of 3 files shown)


ttim:
There is an issue for that: https://github.com/twitter/summingbird/issues/711

ttim:
@johnynek I know here is quite a lot of code without test coverage, if you think it worth it to add tests on that - I can do this. But personally I think it's ok to leave this kind of code without tests.

summingbird-online/src/main/scala/com/twitter/summingbird/online/option/AllOpts.scala
```diff
@@ -98,8 +99,22 @@ trait SummerBuilder extends Serializable {
 
 /**
  * The SummerConstructor option, set this instead of CacheSize, AsyncPoolSize, etc.. to provide how to construct the aggregation for this bolt
+ * @see [[Summers]] for useful [[SummerConstructor]]s.
  */
-case class SummerConstructor(get: SummerBuilder)
+case class SummerConstructor(get: SummerConstructorSpec)
+
+trait SummerConstructorSpec {
```
pankajroark: Why is it called SummerConstructorSpec, there's no SummerConstructor in definition.

summingbird-online/src/main/scala/com/twitter/summingbird/online/option/AllOpts.scala
```diff
@@ -98,8 +99,22 @@ trait SummerBuilder extends Serializable {
...
+  def builder(counter: Name => Counter with Incrementor): SummerBuilder
```
pankajroark: tbh, I think we should just change the definition of SummerBuilder to take the counterBuilder as argument and default to a no-op counterBuilder. SummerBuilder is used very rarely and I think it's ok to break it. There are only a couple places that we'll need to fix at Twitter. Seems better than adding a level of indirection here.
@johnynek wdyt?

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/BuildSummer.scala
```diff
@@ -16,120 +16,58 @@
...
-                executor.awaitTermination(10, TimeUnit.SECONDS)
-              }
-            })
-          }
-        }
+        val asyncPoolSize = option(DEFAULT_ASYNC_POOL_SIZE)
+        val valueCombinerCrushSize = option(DEFAULT_VALUE_COMBINER_CACHE_SIZE)
+        val doCompact = option(CompactValues.default)
+        Summers.async(
+          cacheSize, flushFrequency, softMemoryFlush, asyncPoolSize, doCompact, valueCombinerCrushSize
```
pankajroark: too long, move each param to separate line.

summingbird-online/src/main/scala/com/twitter/summingbird/online/option/AllOpts.scala
ttim: After offline discussion renamed it to `SummerWithCountersBuilder`.

summingbird-online/src/main/scala/com/twitter/summingbird/online/option/AllOpts.scala
ttim: I did this change in backward compatible manner - it's still possible to create `SummerWithCountersBuilder` out of plain `SummerBuilder`.

pankajroark (approved):

summingbird-online/src/main/scala/com/twitter/summingbird/online/option/AllOpts.scala
```diff
@@ -98,8 +99,22 @@ trait SummerBuilder extends Serializable {
 
 /**
  * The SummerConstructor option, set this instead of CacheSize, AsyncPoolSize, etc.. to provide how to construct the aggregation for this bolt
+ * @see [[Summers]] for useful [[SummerWithCountersBuilder]]s.
  */
-case class SummerConstructor(get: SummerBuilder)
+case class SummerConstructor(get: SummerWithCountersBuilder)
+
+trait SummerWithCountersBuilder {
+  def create(counter: Name => Counter with Incrementor): SummerBuilder
```
johnynek: I am not crazy about `with` what about `Name => (Counter, Incrementor)` do they have to be bound together?

summingbird-online/src/main/scala/com/twitter/summingbird/online/option/AllOpts.scala
```diff
@@ -98,8 +99,22 @@ trait SummerBuilder extends Serializable {
 
 /**
  * The SummerConstructor option, set this instead of CacheSize, AsyncPoolSize, etc.. to provide how to construct the aggregation for this bolt
+ * @see [[Summers]] for useful [[SummerWithCountersBuilder]]s.
  */
-case class SummerConstructor(get: SummerBuilder)
+case class SummerConstructor(get: SummerWithCountersBuilder)
+
+trait SummerWithCountersBuilder {
```
johnynek: you probably want this to `extend Serializable` or we may have some corner case issues.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/BuildSummer.scala
```diff
@@ -16,120 +16,58 @@
...
-  @transient private val logger = LoggerFactory.getLogger(BuildSummer.getClass)
+  @transient private[storm] val logger = LoggerFactory.getLogger(BuildSummer.getClass)
 
-  def apply(builder: StormTopologyBuilder, node: StormNode) = {
-    val opSummerConstructor = builder.get[SummerConstructor](node).map(_._2)
-    logger.debug(s"Node (${builder.getNodeName(node)}): Queried for SummerConstructor, got $opSummerConstructor")
+  def apply(builder: StormTopologyBuilder, node: StormNode): SummerBuilder = {
+    val summerBuilder = builder.get[SummerConstructor](node)
+      .map { case (_, constructor) => constructor.get }.getOrElse {
+      logger.info(s"[${builder.getNodeName(node)}] use legacy way of getting summer builder")
```
johnynek: can we indent here?

johnynek (commented):
👍 

some minor comments, but it looks like a nice refactor.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/BuildSummer.scala
ttim: Sure

summingbird-online/src/main/scala/com/twitter/summingbird/online/option/AllOpts.scala
ttim: I prefer to keep it this way (or maybe change to just `Incrementor`). It should be the same object, not both of them.

summingbird-online/src/main/scala/com/twitter/summingbird/online/option/AllOpts.scala
ttim: Actually I prefer to keep it non serializable, this code (creating actual `SummerBuilder` with `Counter`s) should happen only on submitter node, and `SummerBuilder` is an object which should be serializable (and it is).

ttim: Commented in code

summingbird-online/src/main/scala/com/twitter/summingbird/online/option/AllOpts.scala
ttim: Changed to just `Incrementor`.

ttim:
@johnynek @pankajroark can you look again did I point all your comments? Thanks!

johnynek:
👍

PR #739, opened 2017-07-01
Fix summer options propagation

ttim:
Currently if you have `flatMap().name('flatMap').sumByKey().name('sumByKey')` and you have some `Summer` options on both `flatMap` and `sumByKey` for both of nodes (for partial aggregation on `flatMap` and for final aggregation on `sumByKey`) only options from `flatMap` will be applied. 

Root cause of this is in a way we get options for `Node`s - we get options for first producer in producers chain corresponding to `Node`. In case of `SummerNode` members contains two `Producer`s - `IdentityKeyedProducer` and `Summer`.

This PR fixes that by passing `Producer` instance to `BuildSummer` which is used to get options for summer.

### DIFF
--- a/summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
+++ b/summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
@@ -1,7 +1,7 @@
 package com.twitter.summingbird.storm
 
 import com.twitter.algebird.Semigroup
-import com.twitter.summingbird.{ IdentityKeyedProducer, LeftJoinedProducer, Options, Summer }
+import com.twitter.summingbird.{ LeftJoinedProducer, Options, Summer }
 import com.twitter.summingbird.batch.{ BatchID, Batcher, Timestamp }
 import com.twitter.summingbird.online.executor
 import com.twitter.summingbird.online.executor.KeyValueShards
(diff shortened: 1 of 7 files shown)


summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
```diff
@@ -309,9 +309,11 @@ private[storm] case class StormTopologyBuilder(options: Map[String, Options], jo
     }
 
   private[storm] def get[T <: AnyRef: ClassTag](node: StormNode): Option[(String, T)] =
-    node.firstProducer.flatMap { producer =>
-      Options.getFirst[T](options, stormDag.producerToPriorityNames(producer))
-    }
+    node.members
+      .find(!_.isInstanceOf[IdentityKeyedProducer[_, _, _]])
```
pankajroark: Can we comment here why this is required?

pankajroark (commented):
Can you update the description of the PR to indicate how including IdentityKeyedProducer in options calculation is an issue and why removing that is the right fix for this issue.

summingbird-storm/src/main/scala/com/twitter/summingbird/storm/StormTopologyBuilder.scala
ttim: I changed a way I fixed the issue. Turned out way with ignoring `IdentityKeyedProducer`s caused more bugs than fixed.

johnynek:
👍

PR #740, opened 2017-07-06
Add performance tests for flatMap operation in OperationContainer

ttim:
This PR adds tests for `flatMap` / `flatMap` composition in context of `OperationContainer`. Tests are implemented using JMH.

Tests are not ideal, but they give us some estimation of costs our abstractions introduce.  
Most interesting overheads: `IntermediateFlatMap` with `FlatMapOperation` introduces about 1s per 1KK operation overhead. `FlatMapOperation` composition adds another ~1-2s per 1KK.


summingbird-online-perf/src/main/scala/com/twitter/summingbird/online/executor/TestUtils.scala
```diff
@@ -0,0 +1,43 @@
...
+        case (states, output) =>
+          states.foreach { _ => left -= 1 }
+          output match {
+            case Success(value) => value.foreach(bh.consume)
+            case Failure(exception) => bh.consume(exception)
+          }
+      }
+    }
+    while (left != 0) {
+      Thread.sleep(TickDelay.toMillis)
```
johnynek, line 30: I'm pretty skeptical we are comparing apples to apples when we compare this to non `OperationContainers`.

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/SimpleFlatMap.scala
```diff
@@ -0,0 +1,11 @@
+package com.twitter.summingbird.online.executor
+import chain.Chain
+import scala.util.{Success, Try}
+
+class SimpleFlatMap[I, O, S](f: I => TraversableOnce[O]) extends OperationContainer[I, O, S] {
+  override def executeTick: TraversableOnce[(Chain[S], Try[TraversableOnce[O]])] =
+    None
+
+  override def execute(state: S,data: I): TraversableOnce[(Chain[S], Try[TraversableOnce[O]])] =
+    Some((Chain.single(state), Success(f(data))))
```
johnynek: you might put `Try(f(data))` here just to catch any issues with the `f` that could exist since we have try anyway.

johnynek (commented):
looks like composition in these operations is costly.

Note, the optimizer rules could have composed these things for us. I don't know why we wouldn't want to use them.

The main issue was that we didn't work out tracking the naming through the compositions correctly before. But if we can use the optimizer rules we can solve these problems once and use it in scalding, memory, concurrent memory, storm, etc...

summingbird-online-perf/src/main/scala/com/twitter/summingbird/online/executor/TestUtils.scala
ttim: I'm skeptical as well, but I wanted to see at least some number for non `OperationContainer` execution. In the end I don't think it should be anyhow actionable.

summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/SimpleFlatMap.scala
ttim: Done

johnynek:
I'm fine with merging this, but should we add a test that the SimpleFlatMap is identical in performance to the IntermediateFlatMap? Someone might use it as is, and currently it is untested.

Is this a candidate for running in our actual planner, or just a limit to shoot for?

ttim:
@johnynek According to tests `SimpleFlatMap` is much more performant than `IntermediateFlatMap` =)
Anyway I moved it to perf module for now (and made it private). Let's see later do we want to use it instead of `IntermediateFlatMap` in special cases.

johnynek:
Sorry, I meant identical in function, I realize it is much faster.

ttim:
@johnynek yeah =) issue is - it's not identical because our `FlatMapOperation` able to handle `leftJoin` in a same way as `flatMap`s, so it covers only simple cases.
I guess it's safe to merge, because it's private from external users for now.

summingbird-online-perf/src/main/scala/com/twitter/summingbird/online/executor/TestUtils.scala
fwbrasil: `Thread.sleep` adds noise to the result. The blackhole provides a method to simulate CPU usage. I think it's `consumeCpu`

summingbird-online-perf/src/main/scala/com/twitter/summingbird/online/executor/ComposedFlatMapPerformance.scala
```diff
@@ -0,0 +1,82 @@
...
+import com.twitter.conversions.time.longToTimeableNumber
+import com.twitter.summingbird.online.FlatMapOperation
+import com.twitter.summingbird.online.option.{ MaxEmitPerExecute, MaxFutureWaitTime, MaxWaitingFutures }
+import java.util.concurrent.TimeUnit
+import org.openjdk.jmh.annotations.{ Benchmark, BenchmarkMode, Mode, OutputTimeUnit, Scope, State }
+import org.openjdk.jmh.infra.Blackhole
+
+/**
+ * Command to run (should be executed from sbt with project summingbird-online-perf):
+ * jmh:run -i 10 -wi 10 -f1 -t1 ComposedFlatMapPerformance
```
fwbrasil, line 12: I suggest adding these options:

`-prof gc` to enable the GC profiler
`-gc` to force a GC between runs. Without it, 10 warmup and actual iterations might not be enough to exclude GC noise
`-foe true` to fail the benchmark if the code fails. I've seen benchmarks that only test that the benchmark is wrong :)
`-jvmArgs "-Xms1G -Xmx1G"` to make sure that the allocated memory is the same across different dev environments

summingbird-online-perf/src/main/scala/com/twitter/summingbird/online/executor/ComposedFlatMapPerformance.scala
```diff
@@ -0,0 +1,82 @@
...
+      MaxWaitingFutures(1000),
+      MaxFutureWaitTime(1000.millis),
+      MaxEmitPerExecute(Int.MaxValue)
+    )
+
+  @Benchmark
+  @BenchmarkMode(Array(Mode.AverageTime))
+  @OutputTimeUnit(TimeUnit.MILLISECONDS)
+  def measureDirect(bh: Blackhole): Unit =
+    input.foreach(composed.apply(_).foreach(bh.consume))
```
fwbrasil, line 57: The function allocation here (`composed.apply(_)`) introduces some noise. Maybe it's worth extracting it to a class val

summingbird-online-perf/src/main/scala/com/twitter/summingbird/online/executor/TestUtils.scala
fwbrasil: Sorry, I've just noticed that this is actually to check when it's completed. Maybe you could use a `CountDownLatch` for it?

summingbird-online-perf/src/main/scala/com/twitter/summingbird/online/executor/TestUtils.scala
```diff
@@ -0,0 +1,43 @@
...
+
+  def testOperationContainer[I, O](
+    container: OperationContainer[I, O, InputState[Int]],
+    inputs: Array[I],
+    bh: Blackhole
+  ): Unit = {
+    var left = inputs.length
+    for (
+      input <- inputs;
+      emitted <- container.execute(InputState[Int](1), input)
```
fwbrasil, line 18: Do you have a sense of how expensive `container.execute` is? I ask because `testOperationContainer` introduces some noise so its overhead could dominate the final result.

johnynek:
@ttim we are starting to really push on summingbird performance.

Any more work here? Would it make sense to have a video sync sometime soon as we are currently investing in summingbird at Stripe (but also considering if that makes sense or to shift investment somewhat).

If we can share effort, that is a big win. If we each are just going to work privately on our forks maybe each should make their own call.

Thoughts @pankajroark ?

pankajroark:
@johnynek Great to hear about Summingbird performance push. Happy to set up a video sync. As you've seen we've made some optimizations and we wanted to make more. Unfortunately we have to focus resources on other projects for the time being, but we plan to come back and make more optimizations. It will be great to learn about what you've planned. Let me sync up with you offline to set up a time for video sync.

johnynek:
@pankajroark sounds good. I replied offline about a meeting time.

PR #752, opened 2018-02-05
avoid name computation unless there is a cache miss

oscar-stripe:
This is eating a ton of time on our big graphs: we always recompute names even if we hit the cache. For graphs that fork and join a bunch, this can make planning in summingbird exponentially complex.

By moving it, we only compute names when needed and only on a cache miss.


oscar-stripe:
cc @ttim @ianoc

oscar-stripe:
cc @non @erik-stripe

ttim (approved):
Looks good!

PR #753, opened 2018-02-08
Improve memoization to deal with diamonds better

johnynek:
In large summingbird graphs, merges can cause exponential runtime in planning because we repeatedly call on the state function on both sides of the merge.

We had a partial solution for this, which worked to make sure the data was not incorrect, but there is a second outer layer: computing what the acceptable time range is. This was the layer that was not memoized, and it was making it take forever to plan for us.

### DIFF
--- a/summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/ScaldingPlatform.scala
+++ b/summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/ScaldingPlatform.scala
@@ -252,6 +252,24 @@ object Scalding {
   def also[L, R](ensure: FlowToPipe[L], result: FlowToPipe[R]): FlowToPipe[R] =
     joinFP(ensure, result).map { case (_, r) => r }
 
+  def memoizePlan[A](pf: PlannerOutput[A]): PlannerOutput[A] = {
+    val memo = new Memo[FactoryInput, (FactoryInput, A)]
+    StateWithError
+      .getState
+      .flatMap { input =>
+        memo.get(input) match {
+          case None =>
+            for {
+              a <- pf
+              state <- StateWithError.getState[FactoryInput]
+              _ = memo.set(input, (state, a))
+            } yield a
+          case Some((out, a)) =>
+            StateWithError.putState(out).map(_ => a)
+        }
+      }
+  }
+
   /**
    * Memoize the inner reader
    *  This is not a performance optimization, but a correctness one applicable
@@ -262,10 +280,20 @@ object Scalding {
    *  If we memoize with this function, it guarantees that the PipeFactory
    *  is idempotent.
    */
-  def memoize[T](pf: PipeFactory[T]): PipeFactory[T] = {
-    val memo = new Memo[T]
-    pf.map { rdr =>
-      Reader({ i => memo.getOrElseUpdate(i, rdr) })
+  def memoize[T](pf: PlannerOutput[FlowProducer[T]]): PlannerOutput[FlowProducer[T]] = {
+    val memo = new Memo[FlowInput, T]
+    memoizePlan(pf).map { rdr =>
+      Reader[FlowInput, Option[T]](memo.get(_))
+        .flatMap {
+          case None =>
+            for {
+              input <- Reader[FlowInput, FlowInput](fi => fi)
+              t <- rdr
+              _ = memo.set(input, t)
+            } yield t
+          case Some(t) =>
+            Reader.const(t)
+        }
     }
   }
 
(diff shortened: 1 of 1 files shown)


johnynek:
@ttim can you take a look?

ttim:
@johnynek I'll take a look later today!

summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/ScaldingPlatform.scala
```diff
@@ -262,10 +280,20 @@ object Scalding {
    *  If we memoize with this function, it guarantees that the PipeFactory
    *  is idempotent.
    */
-  def memoize[T](pf: PipeFactory[T]): PipeFactory[T] = {
-    val memo = new Memo[T]
-    pf.map { rdr =>
-      Reader({ i => memo.getOrElseUpdate(i, rdr) })
+  def memoize[T](pf: PlannerOutput[FlowProducer[T]]): PlannerOutput[FlowProducer[T]] = {
```
ttim, line 287: Can we put `PipeFactor[T]` back here?

summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/ScaldingPlatform.scala
```diff
@@ -252,6 +252,24 @@ object Scalding {
   def also[L, R](ensure: FlowToPipe[L], result: FlowToPipe[R]): FlowToPipe[R] =
     joinFP(ensure, result).map { case (_, r) => r }
 
+  def memoizePlan[A](pf: PlannerOutput[A]): PlannerOutput[A] = {
```
ttim: I think generic `memoize(StateWithError)` is simpler to understand.

ttim: This function doesn't memoize failures?

summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/ScaldingPlatform.scala
```diff
@@ -262,10 +280,20 @@ object Scalding {
    *  If we memoize with this function, it guarantees that the PipeFactory
    *  is idempotent.
    */
-  def memoize[T](pf: PipeFactory[T]): PipeFactory[T] = {
-    val memo = new Memo[T]
-    pf.map { rdr =>
-      Reader({ i => memo.getOrElseUpdate(i, rdr) })
+  def memoize[T](pf: PlannerOutput[FlowProducer[T]]): PlannerOutput[FlowProducer[T]] = {
+    val memo = new Memo[FlowInput, T]
+    memoizePlan(pf).map { rdr =>
```
ttim: Is `memoizePlan(pf).map { ... as before ... }` has different semantic with this code?

summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/ScaldingPlatform.scala
```diff
@@ -649,11 +677,22 @@ object Scalding {
...
 
-// Jank to get around serialization issues
-class Memo[T] extends java.io.Serializable {
-  @transient private val mmap = scala.collection.mutable.Map[(FlowDef, Mode), TimedPipe[T]]()
-  def getOrElseUpdate(in: (FlowDef, Mode), rdr: Reader[(FlowDef, Mode), TimedPipe[T]]): TimedPipe[T] =
-    mmap.getOrElseUpdate(in, rdr(in))
+class Memo[A, B] extends java.io.Serializable {
+  // Jank to get around serialization issues
+  @transient private[this] var mmap = Map.empty[A, B]
+  def getOrElseUpdate(a: A, b: => B): B =
```
ttim: Can be removed

ttim (changes requested):
Unrelated to your change I'm not quite understand comment before `memoize` function. It says you need to do this memoization for correctness but I can't see how. Do we build `PipeFactory` more than once?

summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/ScaldingPlatform.scala
johnynek: It was needlessly constraining the type. This works for any T. `PipeFactory[T] == PlannerOutput[FlowProducer[TimedPipe[T]]]` so I didn't want to limit it.

summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/ScaldingPlatform.scala
johnynek: it does not memoize failures.

summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/ScaldingPlatform.scala
johnynek: Previously, we didn't memoize the state computation, just the reader inside. The semantics are the same, but this removes an exponential blowup on graphs that merge back into themselves repeatedly (similar to the problem with a naive implementation of fibonacci).

summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/ScaldingPlatform.scala
ttim: Makes sense, looks better this way.

summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/ScaldingPlatform.scala
ttim: Oh, I see. Basically you do `memoizeReader(Reader[I, O]) -> Reader[I, O]`. Then I prefer to have it extracted and then what you want is `memoizeState(pf).map(memoizeReader)`.

johnynek:
@ttim I made the changes you requested. Can you take another look?

ttim (approved):
I think with `memoizeReader` this would be better but I'm good either way.

johnynek:
Thanks for the review @ttim

PR #754, opened 2018-02-08
memoize some Producer hashCodes

johnynek:
This is the same class of bugfix as:

https://github.com/stripe/dagon/pull/19
https://github.com/twitter/scalding/pull/1789


non:
LGTM

ttim:
Looks good but I would be happy if there any way to put this memoization on the top level `Producer` trait.

ttim (approved):

johnynek:
@ttim if you can think of a way, let me know. This fixed one class of exponential time bugs we were seeing

PR #756, opened 2018-02-13
Improve Producer equals methods

johnynek:
This is the same fix as here:

https://github.com/twitter/scalding/pull/1798

This fix is import for us due to some large graphs we have been constructing.


johnynek:
quick look @ttim ?

non:
This looks reasonable to me. 👍

PR #758, opened 2018-02-14
Use dagon for dag rewrites

johnynek:
Scalding 0.18 is using dagon to implement its new backend. We might as well use dagon here too, which was taken out from summingbird about 6 months or so ago.

We have found a number of performance improvements in the code since then, so it should be a win.


johnynek:
@ttim @erik-stripe @non ptal

johnynek:
note we ran these rules in production on storm and significantly reduced the number of storm nodes we created.

I think you should consider using these to optimize your storm graphs as well.

non:
Yeah I can second this recommendation. 👍

johnynek:
These tests are in a very bad place.

I have been clicking rebuild to deal with storm or hadoop timeouts for about 24 hours now. Each trial is another 20-60 minutes. I am almost ready to merge without the tests...

We need to really think about improving the platform tests.

PR #759, opened 2018-02-15
merge in cherry pick of dagon use

johnynek:
0.10 version of #758


johnynek:
looks like some code does not compile on 2.10.... oh well. probably not worth fixing....

maybe just drop 2.10 on 0.10 like we did for 0.11

PR #761, opened 2018-02-28
clean up some closures

johnynek:
This fixes a closure that was causing a problem: FlowDef is not serializable, so capturing it means you have to use kryo, but with scalding 0.18, somehow kryo was hitting what looked like a bug with kryo 2.21 (type mismatch at deserialization time).

ptal @ttim @fwbrasil

### DIFF
--- a/summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/batch/BatchedStore.scala
+++ b/summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/batch/BatchedStore.scala
@@ -48,6 +48,14 @@ case class LTuple2[T, U](_1: T, _2: U) {
 
 }
 
+case class WriteFilter[K, V](batchID: BatchID, latestTime: Timestamp, pruning: PrunedSpace[(K, V)]) extends Function1[(BatchID, (K, V)), Boolean] {
+  def apply(bkv: (BatchID, (K, V))) = {
+    val b = bkv._1
+    val kv = bkv._2
+    (b == batchID) && !pruning.prune(kv, latestTime)
+  }
+}
+
 trait BatchedStore[K, V] extends scalding.Store[K, V] { self =>
   /** The batcher for this store */
   def batcher: Batcher
@@ -96,10 +104,9 @@ trait BatchedStore[K, V] extends scalding.Store[K, V] { self =>
         // make sure we checkpoint to disk to avoid double computation:
         val checked = if (batches.size > 1) lastVals.forceToDisk else lastVals
         batches.foreach { batchID =>
-          val thisBatch = checked.filter {
-            case (b, kv) =>
-              (b == batchID) && !pruning.prune(kv, batcher.latestTimeOf(b))
-          }
+          // closures are the devil
+          val filterFn = WriteFilter(batchID, batcher.latestTimeOf(batchID), pruning)
+          val thisBatch = checked.filter(filterFn)
           writeLast(batchID, thisBatch.values)(flow, mode)
         }
     }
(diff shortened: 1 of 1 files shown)


summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/batch/BatchedStore.scala
```diff
@@ -96,10 +104,9 @@ trait BatchedStore[K, V] extends scalding.Store[K, V] { self =>
         // make sure we checkpoint to disk to avoid double computation:
         val checked = if (batches.size > 1) lastVals.forceToDisk else lastVals
         batches.foreach { batchID =>
-          val thisBatch = checked.filter {
-            case (b, kv) =>
-              (b == batchID) && !pruning.prune(kv, batcher.latestTimeOf(b))
-          }
+          // closures are the devil
+          val filterFn = WriteFilter(batchID, batcher.latestTimeOf(batchID), pruning)
```
fwbrasil, line 108: Curiosity: why create a new class instead of isolating captured values using local `val`s?

summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/batch/BatchedStore.scala
```diff
@@ -204,26 +211,24 @@ trait BatchedStore[K, V] extends scalding.Store[K, V] { self =>
         .toTypedPipe
     }
 
-    /**
```
fwbrasil, line 207: is this just a refactoring or does it fix serialization issues as well?

summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/batch/BatchedStore.scala
johnynek: just to be 100% clear about what is captured. a side benefit is that is has a good equality, but I doubt that will come into play.

I was surprised to see what was captured here because you have 3 nested layers of closure.

summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/batch/BatchedStore.scala
johnynek: it does! see how tricky it is! without this change, you serialize the whole store to get flatOpt, when really we don't need to grab that function for two simple use cases. Not even sure there is no flatten on Option still in scala, but in any case, flatten is not the best call in my view.

summingbird-scalding/src/main/scala/com/twitter/summingbird/scalding/batch/BatchedStore.scala
fwbrasil: 👍 thanks for the clarification. I think we could do better regarding capturing. We could do something at runtime like Spark does or even try to inspect the function using a macro to detect captured values that aren't serializable.

johnynek: given that lightbend are trying to remove macros, or at least drastically narrow the scope, I'm a little worried about being too macro reliant. Note that scalacenter's spores are not even available on 2.12 due to issues with macro interaction with a new version.

fwbrasil:
👍 LGTM (note that one of the ci jobs failed)

johnynek:
sadly this ci is very noisy. scalding often times out, and storm maybe 1/2 the time fails the laws with what we believe to be duplication or loss since the machines get so loaded. it is not a good situation.

PR #762, opened 2018-03-09
call cascading backend to plan writes

fwbrasil:
@ttim This is required for summingbird to work with the new scalding version


fwbrasil:
the CI won't pass since the branch has a dependency to an older scaling version. I can't change the version to use a new release because we're releasing scalding based on a branch. I tested this patch using the sandbox and it's ok internally cc/ @pankajroark

johnynek:
we could publish a `0.18.0-RC1` so we could keep the CI here green.

pankajroark:
@johnynek This branch is just for internal testing right now and we are ok with ci failing for this branch. Let me merge this for now.

pankajroark:
This is just a personal branch, we can revert easily if there is a reason.

PR #764, opened 2018-04-09
Make mergeBatched protected instead of private.

non:
There are reasons why particular implementations might want to
override this method (for example, to support custom serialization
strategies), so let's make it protected instead of private.


johnynek:
👍 

Thanks!

merge when green.

johnynek:
looks like this fails mima, naturally... 

So we can make the change at the next mima break, maybe when we release for scalding 0.18

PR #766, opened 2018-07-02
Externalize Monoid in flat map and summer bolts

ttim:
Currently `Monoid` instance passed to `sumByKey` is serialized through java serialization (for summingbird-storm platform), while other user's code (e.g. flatMap functions) is serialized through `Externalizer` (which tries java serialization and kryo in case it didn't work). Seems like this behaviour should be consistent and we should serialize `Monoid` through `Externalizer` as well.

In this PR I've added test for this and fixed it via putting `Monoid` into `Externalizer` box.

### DIFF
--- a/summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/FinalFlatMap.scala
+++ b/summingbird-online/src/main/scala/com/twitter/summingbird/online/executor/FinalFlatMap.scala
@@ -51,7 +51,8 @@ private[summingbird] case class KeyValueShards(get: Int) {
     math.abs(k.hashCode % get)
 }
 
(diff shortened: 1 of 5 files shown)


johnynek (approved):
Seems good. It would be good to have a test for each part of the change such that it fails without it, but if we don’t have time I’m not opposed to merging.

Thanks for finding this issue.

ttim:
> It would be good to have a test for each part of the change such that it fails without it, but if we don’t have time I’m not opposed to merging.

When I was testing firstly I ran test without changes (failed with `Tail` bolt), then with only summer change (failed with `FlatMap-Tail` bolt) and then with both changes. Since this test covers both partial aggregation and full aggregation I think it's enough.

PR #767, opened 2018-07-28
Configure Cirrus CI

fkorotkov:
Cirrus CI allows to use more than 2 CPUs which speeds up tests drastically:

image

Cirrus CI also provides more flexibility by using containers and very fast caching storage.

In case you'll decide to merge, an admin of Twitter organization must install Cirrus CI App from GitHub Marketplace: https://github.com/marketplace/cirrus-ci

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,0 +1,18 @@
+container:
+  image: bigtruedata/sbt:latest
+  cpu: 4
+  memory: 12G
+
+test_task:
+  env:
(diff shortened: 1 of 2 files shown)


johnynek:
This is interesting! Thanks for sending the PR. 

We need to ask Remy or maybe Timur can help us find the right folks.

fkorotkov:
Sure thing! @ttim was complaining about unstable CI so I took a look 😅

PR #771, opened 2019-01-14
Add `modifyBeforeWrite` extention point in `VersionedBatchStore`

ttim:
In order to better support different GDPR use cases in Twitter we need to have a way to do modification for stored `(K, V)` pairs in Summingbird's Scalding platform.

In this PR I've added `modifyBeforeWrite` protected method which suits this needs in `VersionedBatchStore`.


dieu (approved):

ttim:
@johnynek are you ok with this change?

johnynek:
totally fine with me.

Sorry for the delay.

Our data is small enough that we just recompute everything every so many days, I know this isn't practical for you.

PR #774, opened 2019-02-11
Avoid deprecated com.twitter.conversions

kevinoliver:
Problem

Twitter's util deprecated `com.twitter.conversions.time` in https://github.com/twitter/util/commit/c4c33a3424c53cdad13b05536abfaa07eb78ff22.

Solution

Use `com.twitter.conversions.DurationOps`.

PR #775, opened 2019-02-11
Completely avoid util conversions

kevinoliver:
Problem

As noted in https://github.com/twitter/summingbird/pull/774 Twitter util deprecated some conversions. However, summingbird is on an older version of util that does not have the new APIs.

Solution

Avoid the conversions sugar completely.


ttim (approved):
Thanks!

I'll merge it as soon as CI is green.

